PR #3, opened 2023-03-13
add zh_Hans localization

wfjsw:
https://github.com/wfjsw/stable-diffusion-webui-localization-zh_Hans

### DIFF
--- a/index.json
+++ b/index.json
@@ -873,6 +873,13 @@
             "description": "Old unmaintained localizations that used to be a part of main repository",
             "added": "2022-11-08",
             "tags": ["localization"]
+        },
+        {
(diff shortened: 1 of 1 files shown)


dtlnor:
recommanding region specific localization name https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/3348#issuecomment-1290421450

wfjsw:
Since there is no usage difference on Simplified Chinese between Mainland China and Singapore, it is more appropriate to use `zh_Hans` to signal that this can be used in both region. Also `zh_CN` is incompatible with the translation platform and the continuous integration engine currently in use.

PR #15, opened 2023-03-27
Added Prompt Translator extension

ParisNeo:
Added Prompt translation extension to the list of extensions

### DIFF
--- a/index.json
+++ b/index.json
@@ -21,6 +21,13 @@
     },
     "extensions": [
         {
+            "name": "Prompt Translator",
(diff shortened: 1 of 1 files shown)


vladmandic:
I'm not sure you want to have localization tag - that is used exclusively for UI language packs and typically not even displayed by default when listing extensions.

I get why you set it, just noting.

ParisNeo:
Ok, I'll remove it

ParisNeo:
The tag has been removed.
Thank you

PR #24, opened 2023-04-07
change the org of zh-hans localization

wfjsw:


w-e-w:
btw, it seems github automatically redirects the old wfjsw url to hanamizuki-ai

wfjsw:
yeah I know but it is implicit and unreliable tho

PR #25, opened 2023-04-07
Fix sd-discord-rich_presence url typo

harukaxxxx:
fix sd-discord-rich_presence url


w-e-w:
thanks
how did I not catch that

PR #28, opened 2023-04-13
add feed extension

todhm:
Hello!
The reason why Feed extension is necessary for Web UI is as follows.

I have been using Web UI since January. In Korea, there is a community called “Arca.live” where AI images are actively shared. I discovered a unique user loop:

1.Browse AI images in the community.
2.Find AI images you want to recreate in the community.
3.Create your own image by referring to the parameters of the image you want to recreate in Web UI.
4.Show off well-made images in the community.

I believe that the value of AI images lies in browsing, recreating, and showing off.
To access this community directly within Web UI, I developed a Feed extension, which requires using an external server to save images.

Web UI is a creation tool, but if community features are added, it would be a great help in building the user loop or ecosystem I mentioned earlier. Image producers can gain reputation and potentially generate revenue, while image consumers can easily browse images and use them as inputs for the Web UI tool.

### DIFF
--- a/index.json
+++ b/index.json
@@ -978,6 +978,13 @@
             "description": "Old unmaintained localizations that used to be a part of main repository",
             "added": "2022-11-08",
             "tags": ["localization"]
+        },
+        {
+            "name": "WebUI Feed Extension",
+            "url": "https://github.com/todhm/sd_feed.git",
+            "description": "Share your images with the community and receive reputation based on the feedback you receive from other",
+            "added": "2023-04-13",
+            "tags": ["online", "tab"]
         }
     ]
 }


w-e-w:
do you realize not having any sort of authentication is a bad idea?
the user is purely tracked by a UUID
and you can easily read other people's UUID
which means anyone can impersonate anyone

todhm:
Dear w-e-w 
I understand your concern. However, we don't expose the user's UUID to other users, and since we don't expose other users' UUID to the public it is hard to hack other's information. 
However, to reduce your concern we added token based auth system so it's getting harder to use others' information. I hope you understand that anyone can impersonate others' information. 
Also please understands that we make users to understand that their image information shared through other users.

w-e-w:
I, who is not a security expert at any type is able to defeat your system in 5 minutes
you can easily retrieve the uploaders users UUID by looking at the image URL

giving that your team is unaware of this exploit
I take this as proof that your team is incapable of making a secure application on a very basic level

As such I again will not be approving this
(until the security vulnerabilities are addressed)

if you do submit the extension again, due to the nature of this extension and the track record
the review should be done not only by me alone
I think it's a good idea for more people to review and approve of this before this gets added

screenshot proof of the vulnerability exist
image
image
image
needless to say that I'm not these two users

to make it clear 
my personal feelings about this is positive
I like the concept and I think it could be beneficial
please don't take my rejection as a person attack to your / your project or your team
I would like to approve it and allowing more users to be aware of this extension
but unfortunately, due to the lack of security awareness I cannot in good consciousness approve this

other possible vulnerabilities that I'm suspecting might exist
malicious actors might upload files to your server
and when users run webui, it might be automatically downloaded to the browser
I can see this as a potential avenue for exploit
like I said before I'm no security expert I'm not sure if such an issue exists
but my gut feeling tells me that this is an issue

some other concerns to name a few
by submitting images to your service who owns the rights to the image?
if a user wish to delete a image can they?
what about deleting messages
who is responsible of the service if something goes wrong
are you going to be running ads in the future?
are you possibly tracking users based on the uuid and IP relative location?
why are you sending all those image click messages?
I failed to see the point of making this a user-based service rather than an anonymous one
they only benefits is that user can communicate with each other and possibly from friends
and possibly following certain users
with how the service is currently set up it works with small amounts of users but with more users I don't see this system being useful
with more users there's hardly any point of keeping track of individual IDs

a anonymous system could be far more easy to implement
and far less of a security concern
the basic model of how anonymous sites work is that one user can make a post
they are given a special secret key that they can use to delete or edit the post
only the user who have the key can edit or delete the post
with a custom client it's easy to keep save all the posts one have made it in the past on client side
similarly it's also easy to make a favorite list
most of the features that I think would be useful for this type of community can be done anonymously
this also removes the need of a server storing user data with an account  
and less concern of tracking, connecting IP addresses with a user ID

allowing others people to host different image host server is it good idea
as opposed to the one central server you have now, if more people were to use this
I can see that the server will be crippled

todhm:
Thanks for your detailed feedback

Now we hide all the uuid information that can be accessed from api response
As you can see we add authorization token to get authorization of APIS.
However since old users might not have token if they don't download new versions Old users are excepted for auth check.
We will notice them after approval.
스크린샷 2023-04-14 오후 1 58 35


Seconly i will answer your questions.

Q. Some other concerns to name a few by submitting images to your service who owns the rights to the image?
A. Of course, the ownership of the photo belongs to the uploader of the photo.

Q. If a user wish to delete a image can they?
A. If requested by the user, of course, they can delete the photo. In the future, a function to report and delete photos will be added.

Q. what about deleting messages? 
A. If requested by the user, of course, they can delete the photo. In the future, a function to report and delete photos will be added.

Q. Who is responsible of the service if something goes wrong?
A. Me!

Q. Are you going to be running ads in the future?
A. Commercial advertising is not my ultimate goal. I dream of creating an ecosystem between AI image producers and consumers.

Q. Are you possibly tracking users based on the uuid and IP relative location?
A. We are not tracking ip. We only tracking users based on the uuid.

Q. Why are you sending all those image click messages?
A. This is because, as mentioned earlier, there is a circular structure in which WebUI is meaningful. Browsing, imitating, and showcased for fame and recognition between producers and consumers of these images. We display the number of image clicks, recommendations, and comments to give pride to the image producers.
The eye icon in the picture below represents the number of clicks/views.

image

ICYDev:
Personally I feel like this 'extension' might be better off as it's own browsing website or something for now. The fact that this extension would be sending messages to some remote server by relatively passive activity doesn't really seem fitting for the 'local nature' of the webui (personal opinion).

Prompt copying and sharing also aren't complicated 'workflows' to begin with, so this extension overall seems like it would likely cause more trouble than it's worth.

It also seems quite possible that there are more security problems in this system that might not yet to have been noticed.
As nice of an idea this extension seem to promise, there just seems to be too much potential issues down the line for this extension.

The same browse-iterate-giveback behavior loop exists in many places other than just ArcaLive (Reddit and Discord are two obvious examples). If there's an extension that can say, browse images from multiple sources, then maybe it would be a worthwhile extension.
I'm sure this extension may one day become very useful for people who enjoys the 'cycle' a lot, but as it currently stands, a browser for a seemingly relatively young and rudimentary image server just doesn't seem worth it for the issues it seems to potentially have. And even if it were to be added to the extensions list, it might need to include some form of warning in the description to inform people of it's potential issues.

w-e-w:
explain to me this
WAY ARE ALL TOKENS THE SAME
this is literally the first test I did, and I feel like you have no idea what the is token is
```py
# created new user
import requests
import uuid

if __name__ == '__main__':
    for _ in range(5):
        userId = str(uuid.uuid4())
        response = requests.post(
            'https://newtypev3-server-vjiloyvjvq-an.a.run.app/user/user',
            json={'userId': userId, 'useAuth': True}
        )
        print(userId, response.json())
```
output
```sh
091a481f-2ab0-47a2-9bf9-7f1347513f37 {'msg': 'OK', 'token': 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.hKgepQxWxC7RQC5cOEf0U-IBeK9tVxsPZS2kpVybgQU'}
2ae94442-8d6f-41b4-ba7e-32836cbb48a9 {'msg': 'OK', 'token': 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.hKgepQxWxC7RQC5cOEf0U-IBeK9tVxsPZS2kpVybgQU'}
750f3ccd-dbfa-4653-a8d6-86bbc522428c {'msg': 'OK', 'token': 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.hKgepQxWxC7RQC5cOEf0U-IBeK9tVxsPZS2kpVybgQU'}
5f5e79dc-49a6-4ca7-ac6a-a7bd825b2d3f {'msg': 'OK', 'token': 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.hKgepQxWxC7RQC5cOEf0U-IBeK9tVxsPZS2kpVybgQU'}
1e88d38c-5412-4fbe-a427-4dc5f68addd4 {'msg': 'OK', 'token': 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.hKgepQxWxC7RQC5cOEf0U-IBeK9tVxsPZS2kpVybgQU'}
```

while you seem to have at least change the server so that it does not expose the ID in the image URL
I cannot be certain if this is the case throughout the entire system
if it's indeed the case that the uuid is never exposed, it is still NOT ideal but can work

from my understanding the reason why most services have a username and password separately in two strings is that it makes it easier to manage
the username / uuid  may be used for tracking users across the site the site in multiple places, and not treated with such care
while the password / token is use and ONLY for authentication
a programmer will be especially careful when handling password / tokens
> I feel like I have to say this "the password should never stored on the server, only store the seeded hash of the password, this is standard security

at this point without seeing the server open source, I won't be approving this, I simply have no trust in your ability of making a secure system
and if this is going to be approved it will have to be vetted not just by me
hopefully by someone more adept in web dev and security
> I'll ask other devs on discord see if someone is up to the task

mcmonkey4eva:
`hopefully by someone more adept in web dev and security` hi that's me

----

You just googled `web token`, clicked the first link, didn't notice that what you clicked was the wrong thing, and then implemented the thing you found incorrectly.
The "token" there is a JSON Web Token, which is a _message signing_ algorithm not a token generator, which you signed the payload `{}` which appears to be an empty JSON string? Resulting in something that vaguely resembles cyphertext well enough to trick somebody inexperienced, but with no actual technical value in any direction. If you actually populated the JSON payload, it would at least generate unique data, but would still not be an actual token.

If you want to implement a webservice with the concept of user accounts, you -must- spend a large amount of time studying and learning about web security, or hire somebody who already knows. It is not something you can slap together by copypasting the first result from a google search, or asking ChatGPT about it.

-----

Your extension is also arbitrarily dumping large amounts of metadata to an unknown remote server for which you have not seemingly published the source to.

Your GitHub account has a history of working on attempts to commercialize webapps like this, alongside cryptocurrency and other finance-centric projects.

Both of these things (privacy-violation through metadata broadcast, and monetization of users) are antithetical to the goals of the Automatic Stable-Diffusion-WebUI. We've blocked out extensions for simple things like advertising before.

-----

The biggest issue _as applies to making an extension_, is what @ICYDev wrote - this has no function being integrated to the WebUI, vs simply being its own website.
This is just a website. No part of it *needs the webui* to function.

mcmonkey4eva:
Those values are unique, but they are still JSON Web Tokens and not actually tokens. As I said previously, web security should not be developed by somebody who has spent zero time researching web security.

This extension also continues to just be a website that has no reason to integrate into the WebUI. Please just make a website.

w-e-w:
really?????
"security through obscurity"
I have no words

papuSpartan:
> Security by obscurity is not secure.

Couldn't have said it better.

w-e-w:
open sourcing your backend is basically your only redeeming chance of convincing people that this is a good idea

and as other have pointed out, the extension literally a website
your feeds tab would literally run fine in the browser with the webui backend not running

w-e-w:
I'm closing this
reasons
1. this project is full with security holes, and made by someone that has no concept of security
2. unwilling to open source they serve a back end, citing the main reason beeing that it might expose security holes
3. not able to prove that the service is safe to use, and whether or not things are done ethically
4. lots of calls to external server which raises security concerns
5. feedback parts of the community that this should just be a standalone website and does not need to be an extension
6. lots of features such as deletion of images it's only promised but not implemented

this does not ban this extension or the service for functioning, if people choose to use your extension, they still and allwase can install it manually

in my opinion
if you are able to grow your extension/service to a substantial user base and convince the community that this needs to be in an extension, you can submit it for reconsideration in the future

PR #31, opened 2023-04-16
Add Infinite image browsing

zanllp:


w-e-w:
it seems this also have integration with baidu's cloud stroage
so the description should make it clear that it also has "option" to integration with baidu
if I'm understanding correctly the dependency for baidu's are separately installed if the use decides to use it
so it shouldn't be a security concern 
but still it may be appropriate to add the `online` tag 

note: I haven't make sure whether or not if it calls to an external server

this extension also raise some concerns especially for those people that are sharing their webui install with others
as the extension have the ability to browse any file on the computer within the browser 
it essentially is file / image browser inside a web page

personal opinion
I wouldn't use this, I would just use File Explorer or some other image browsing software such as XnView

zanllp:
1. This plugin also supports "send to" and is currently developing inverted indexes for prompt, Lora, model, and manual tagging.
image

2. "Infinite" means not only no need for paging, but also the ability to browse images similar to "os.walk" in Python, where it automatically loads the data from the next folder after you finish browsing the current folder.

3. Baidu Cloud is optional.
image

w-e-w:
> More importantly, it provides faster preview speed. My inspiration came from my friends complaining about how slow it was to preview images on Colab or cloud GPU servers.

good point, my mind blanked I totally forgot about cloud gpus

w-e-w:
I personally don't have any issues with it
I will merge this in 24 hours if no one has any objections

w-e-w:
I will suggest you change the internal path of `baidu_netdisk`
someone got extremely perturbed seeing the `http://127.0.0.1:7860/baidu_netdisk/.......` calls, thinking that they somehow have Baidu installed on the local system

I believe the reason why the name `baidu_netdisk` is used was because the project was original called `stable-diffusion-webui-baidu-netdisk`?

zanllp:
> I will suggest you change the internal path of `baidu_netdisk` someone got extremely perturbed seeing the `http://127.0.0.1:7860/baidu_netdisk/.......` calls, thinking that they somehow have Baidu installed on the local system
> 
> I believe the reason why the name `baidu_netdisk` is used was because the project was original called `stable-diffusion-webui-baidu-netdisk`?

I also have this idea of gradually removing unnecessary "Baidu Netdisk", but currently there are higher priority features waiting for me to implement and I don't have any extra time. 

Yes, the earliest function of this project was simply to upload files to Baidu Netdisk. However, later they wanted to preview images in the local file list, and as the functionality continued to improve, it gradually became what it is now.

PR #33, opened 2023-04-17
Adding tag "prompt translation"

nightgolfer:
Proposal for adding a category/tag for extensions that are primarily used for translating prompts from one language to another.

Added tag description, and added "prompt translation" tag to "auto translate" and "prompt translator" extensions.


w-e-w:
at present I don't think it warrants new category

PR #42, opened 2023-05-05
Add SD Model Organizer

alexandersokol:
Hello!
Just finished creating a new extension and would like to add it to the general index.

The main goal is to add some user-friendly UI to manage information about models, add own notes and download/remove files.

You can find all information in the README file in the repo.
https://github.com/alexandersokol/sd-model-organizer

### DIFF
--- a/index.json
+++ b/index.json
@@ -999,6 +999,13 @@
             "description": "Old unmaintained localizations that used to be a part of main repository",
             "added": "2022-11-08",
             "tags": ["localization"]
+        },
+        {
+            "name": "Model Organizer",
+            "url": "https://github.com/alexandersokol/sd-model-organizer",
+            "description": "Extension that allows to store information about different models needed for Stable Diffusion WebUI, add information, description, notes and download them.",
+            "added": "2023-05-05",
+            "tags": ["tab", "UI related"]
         }
     ]
 }
--- a/index.json
+++ b/index.json
@@ -1003,9 +1003,9 @@
         {
             "name": "Model Organizer",
             "url": "https://github.com/alexandersokol/sd-model-organizer.git",
-            "description": "Extension that allows to store information about different models needed for Stable Diffusion WebUI, add information, description, notes and download them.",
+            "description": "Allows to store and manage own model collections, add information, notes, previews, download links of model. Easily download from any external direct link or shared file link from Google Drive. Share own models collection with json file or use remote Firestore database to make it available and synced on several devices.",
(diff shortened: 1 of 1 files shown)


w-e-w:
some clarification and issues I found
clarification 
```
even though you read me makes it seems like so, this is not a civitai integration
this is a extension for downloading models from **any external url** with optional preview image based on entries made by the user manually
with option to fill in other information such as preview image, original model web page and  any other notes that the user would like to add
supports for searching your models by grouping and tags Etc
```
is this correct

some issues I found

1. the preview image doesn't seem to be locally cached, which means it will need to make request to external server to fetch image causing unnecessary strain
2. you seem to be using the wrong path for the models
it is common practice to change the ckpt path 
see https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Command-Line-Arguments-and-Settings
`--ckpt-dir`
there are also many other parts not just the sd checkpoint
3. after I have downloaded file there's an error saying that it's being used by some other process or something
however I don't think this is an issue, since the model is moved to the `internal model folder` `wrong location`

this extension is broken
the extension will have to be working in order for it to be added. and I consider this to be broken

alexandersokol:
Hi!
Yes, your clarification is correct, I might update readme then, because all sample models there from civitai and users might be confused. 
1. Will update it
2. Will check it.
3. Can you share what OS are you tested on? Looks like it is something system-specific. I will check it too.

Thank you for the review!

w-e-w:
windwos 11
chrome
python: 3.10.6  •  torch: 2.0.0+cu118  •  xformers: 0.0.17  •  gradio: 3.28.1  •  commit: [5ab7f213](https://github.com/AUTOMATIC1111/stable-diffusion-webui/commit/5ab7f213bec2f816f9c5644becb32eb72c8ffb89) 

`--ckpt-dir drive:\some\path`
I only tried downloading one sd checkpoint
but I assume other types of models will have similar behavior

alexandersokol:
Hello!

- Updated extension description.
- Added "online" tag.
- Fixed download error.
- Command arguments like `--ckpt-dir`, `--vae-dir`, `--lora-dir`, `--hypernetwork_dir`, `--lyco_dir`, `--embeddings_dir` now have top priority of defining models destination directory in case nothing else was set manually in settings menu.
- temp files during download are now placed in the destination dir to prevent moving huge files across different disks.

w-e-w:
after a quick test it seems to be working

PR #43, opened 2023-05-05
Add Pixel extension

Leodotpy:
This extension adds various pixel art effects.  Currently supports downscaling, color palette limiting and grayscale effects. No models are required.

### DIFF
--- a/index.json
+++ b/index.json
@@ -447,6 +447,13 @@
             "added": "2022-12-24",
             "tags": ["editing"]
         },
+	{
+            "name": "Pixel",
+            "url": "https://github.com/Leodotpy/sd-pixel.git",
+            "description": "Quickly and easily perform downscaling, color palette limiting, and other useful pixel art effects through the extras tab.",
+            "added": "2023-05-05",
+            "tags": ["editing", "extras"]
+        },
         {
             "name": "Pixelization",
             "url": "https://github.com/AUTOMATIC1111/stable-diffusion-webui-pixelization.git",


w-e-w:
you might want to add that you need may to restart in order for the extension to work

Leodotpy:
Added that tip. Thank you :)

w-e-w:
> (You may need to restart the webui once for the extension to load. This can also be done from the extensions page)

ahhh, about that

```py
Restarting UI...
Running on local URL:  http://127.0.0.1:55555

To create a public link, set `share=True` in `launch()`.
Startup time: 0.5s (create ui: 0.3s).
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\routes.py", line 399, in run_predict
    output = await app.get_blocks().process_api(
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\blocks.py", line 1297, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\blocks.py", line 1133, in preprocess_data
    self.validate_inputs(fn_index, inputs)
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\blocks.py", line 1120, in validate_inputs
    raise ValueError(
ValueError: An event handler didn't receive enough input values (needed: 8, got: 0).
Check if the event handler calls a Javascript function, and make sure its return value is correct.
Wanted inputs:
    [label, textbox, textbox, image, image, image, image, image]
Received inputs:
    []
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\routes.py", line 399, in run_predict
    output = await app.get_blocks().process_api(
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\blocks.py", line 1297, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\blocks.py", line 1133, in preprocess_data
    self.validate_inputs(fn_index, inputs)
  File "B:\Downloads\sd.webui\system\python\lib\site-packages\gradio\blocks.py", line 1120, in validate_inputs
    raise ValueError(
ValueError: An event handler didn't receive enough input values (needed: 8, got: 0).
Check if the event handler calls a Javascript function, and make sure its return value is correct.
Wanted inputs:
    [label, textbox, textbox, image, image, image, image, image]
Received inputs:
    []
```
tldr
webui / gradio don't seem to like it 

I'm not sure if this is actually caused by your extension but it did popped up when I was installing / enable / disable
but it did happen after I installed your extension so
and also if I have your extension enabled and then fresh restart
the error doesn't seem to appear

Leodotpy:
Oh that's strange. For now I will fix the readme to mention that. Then I will look further into it and try to solve it.

w-e-w:
I think it may not be your extension but something wrong on my end

Leodotpy:
To check, I just did a fresh install of the extension of a separate device.  I didn't receive any error messages. All it took was one webui reload via "Apply and Restart UI" to get the sliders and functionality to work in the extras tab.

This leads me to believe that it is something on your end.

ICYDev:
Can confirm it works perfectly fine. Dragging the slider for the greyscale is a bit glitchy tho.

w-e-w:
I've had pretty much confirmed that it is my issue

the issue won't really happen unless you're like me that who have to test lots of extensions

making it hard as your debug is because this happens on the clean install because it has nothing to do with the install
> I was testing your extension on clean install

so naturally when I saw the issue I thought it was your extension that's causing the problem

the real issue is that, I've been testing too many things
which basically mean frequent restarts of webui and losts enable / disabling lots of extensions
this results in me having lots of webui web pages still open in the background but I forgotten
because of the difference in extensions those web pages are launched with different JavaScript loaded

and buy coincidence some extensions combined with other extensions combined with some of those old web UI web pages would randomly spill out errors at random intervals
>I'm extremely uncertain which combinations of which ones cause which issues because just too many factors
no ext or ext A or ext B alonefine, , but with A and B together would cause issues
it's basically impossible to do debug

**anyway the issue was on my end having too many old instances of web pages open
because I've been testing too many stuff**

Leodotpy:
Okay thanks I understand. Also, thank you for all your hard work on all these extensions.

PR #45, opened 2023-05-08
Add sd-webui-prompt-all-in-one

Physton:
Improve the user experience of the prompt/negative prompt input box. It has a more intuitive and powerful input interface that offers features such as automatic translation, history records, and favorites.
[https://github.com/Physton/sd-webui-prompt-all-in-one](https://github.com/Physton/sd-webui-prompt-all-in-one)

### DIFF
--- a/index.json
+++ b/index.json
@@ -1013,6 +1013,13 @@
             "description": "Allows users to create video2video and text2video animations using any SD models as a backbone. Please, make sure that 'sd-webui-controlnet' extension is also installed.",
             "added": "2023-05-06",
             "tags": ["tab", "animation"]
+        },
+        {
+            "name": "sd-webui-prompt-all-in-one",
+            "url": "https://github.com/Physton/sd-webui-prompt-all-in-one",
+            "description": "Improve the user experience of the prompt/negative prompt input box. It has a more intuitive and powerful input interface that offers features such as automatic translation, history records, and favorites.",
+            "added": "2023-05-08",
+            "tags": ["UI related", "prompting", "localization"]
         }
     ]
 }
--- a/index.json
+++ b/index.json
@@ -1033,7 +1033,7 @@
             "url": "https://github.com/Physton/sd-webui-prompt-all-in-one",
             "description": "Improve the user experience of the prompt/negative prompt input box. It has a more intuitive and powerful input interface that offers features such as automatic translation, history records, and favorites.",
             "added": "2023-05-08",
-            "tags": ["UI related", "prompting", "localization"]
+            "tags": ["UI related", "prompting", "ads", "online"]
         }
     ]
 }


w-e-w:
1. due to the use of external translation services this extensions requires the `online` tag
2. remove `localization` tag, this tag specifically used for localization files not for translation extensions
3. includes self-advertisement in extension, requires `ads` tag (the GitHub button)
4. poorly translation integrated

suggestion
1. use the more modern prompt emphasis syntax, ie `(1girl:1.5)` as opposed to `((((1girl))))`
2. by default any extensions with ads tag will be hidden on the this, you have the choice of  removing self advertisement or adding the `ads` tag
3. lots of translation services (even the the first one on the list Google translate) doesn't work
4. language codes used by the extension and language codes used by translation service don't match resulting in error
>if a translator is showing and selectable it should work, or it should not show
6. poor dark theme support, parts of the user interface is illegible

Physton:
> 1. due to the use of external translation services this extensions requires the `online` tag
A: Added.
> 2. remove `localization` tag, this tag specifically used for localization files not for translation extensions
A: Modified.
> 3. includes self-advertisement in extension, requires `ads` tag (the GitHub button)
> A: Added.
> 4. poorly translation integrated
> A: Due to the numerous supported translation interfaces, each with its own language code, continuous updates are needed in future updates. I am currently organizing the data and in the future, we will support translation for most languages worldwide.
> 5. extension stores files outside of extension directory, which would cause issues to the user
> A: The data contains important information such as history, bookmarks, API keys, etc. In some integrated UI launchers, updating the extension can lead to the deletion of data within the extension folder. Therefore, the data needs to be stored elsewhere.
> 
> suggestion
> 
> 1. use the more modern prompt emphasis syntax, ie `(1girl:1.5)` as opposed to `((((1girl))))`
> A: Many optimizations have been made in the latest version. Please see:[https://github.com/Physton/sd-webui-prompt-all-in-one#update-record--%E6%9B%B4%E6%96%B0%E8%AE%B0%E5%BD%95](https://github.com/Physton/sd-webui-prompt-all-in-one#update-record--%E6%9B%B4%E6%96%B0%E8%AE%B0%E5%BD%95)
> 2. by default any extensions with ads tag will be hidden on the this, you have the choice of  removing self advertisement or adding the `ads` tag
> A: An ads tag has been added.
> 3. lots of translation services (even the the first one on the list Google translate) doesn't work
> A: Translation services include both those that require an API key and those that do not. Translation services that do not require an API key use web crawlers to call online website's translation services, and whether they can be used depends on each person's computer and network environment. For example, Google Translate cannot be used in China. Our recommendation to users is to use translation services that require an API key.
> 4. language codes used by the extension and language codes used by translation service don't match resulting in error
> A: I don't understand what specific error you are referring to.
> 
> > if a translator is showing and selectable it should work, or it should not show
>> A: Currently, there is no good way to detect whether each translation interface that does not require an API key can work properly. By providing these free interfaces, we can give users more choices.
> 
> 6. poor dark theme support, parts of the user interface is illegible
> A: We will continue to optimize the UI interface in future updates.

w-e-w:
> language codes used by the extension and language codes used by translation service don't match resulting in error
A: I don't understand what specific error you are referring to.

this is actually the issue that's called certain translator and certain languages don't work
the issue is that you have so many translators, when you select a certain local language
your extension uses a certain language code for that language to communicate with the translator
but the problem is different translators may use different language codes there are different from one that you use in ext
I'm guessing you didn't do throw testing on every translator on every language
> yes this is tedious work

I'm guessing a solution would be to have a conversion table that you can convert the language code that used internally and the language code that is actually used by the translator
I haven't looked in your code to see if you already have this conversion table
example if I select traditional chinese which internally uses the language code zh-TW, the error is unsupported language
you can clearly see that it actually supports zh-tw, the issue is that the TW is lowercase
image
I have not done a comprehensive test to see if other similar issue exists but I was able to find this issue within 5 minutes of using the application which levens me to believe that this is very unpolished


>extension stores files outside of extension directory, which would cause issues to the user
A: The data contains important information such as history, bookmarks, API keys, etc. In some integrated UI launchers, updating the extension can lead to the deletion of data within the extension folder. Therefore, the data needs to be stored elsewhere.

I've seen other extension stores there persistent data within the extension directory
I don't know what launcher you're using, I wasn't even aware there is a launcher
if it launcher would randomly decides to wipe someone's data then once you probably not use that

if you store it outside in the webui root, I can see people that don't know the way around git might accidentally stash
everyday information you stored, as the file you added is will not be ignored in the webui repository

also generally speaking if you wish to store data like that it is better to have a way to customize the directory, either within settings or using launch flags

as you have added `ads` `online` tag t
I can see the extension to be useful for parts of the community even though some of his features particularly translator integration is buggy
hopefully you will be able to fix those issues in future development

PR #49, opened 2023-05-11
Nevysha's Cozy Nest ads tag

w-e-w:


w-e-w:
due to your extension having self-advertisements
image
I have to add the `ads` tag
this will result in your extension not show up on the list by default unless the user decides to show it
image
if you decide to remove self advertisement in the extension you may have the ads tag removed

w-e-w:
@Nevysha
due to your extension having self-advertisements
image
I have to add the `ads` tag
this will result in your extension not show up on the list by default unless the user decides to show it
image
if you decide to remove self advertisement in the extension you may have the ads tag removed

Nevysha:
Not problem with that but just to understand : that's because the 15x15px icon is displayed on the top bar or even if wrapped inside the setting it will be tagged as ads ?

w-e-w:
yes it is not intrusive and out of the way
but the thing is the only ones that are able to add links like that are the ones making UI's extensions
this makes it very unfair
should other extension makers put their self advisements in settings, out of the way?
or maybe do a start finding interesting ways of sneaking in links

if we allow this what it mean that we have every extension having some advertisement somewhere hidden?

w-e-w:
this https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/45 extension randomly decide to add a button in the middle of his new add-in toolbat that links to his extension repo
at the bottom of the readme theres Donate info

there's no reason of putting a button therem, one might argue it's for documentation usage manual but I still think it's the advertisement
there's no reason for the link to be there in the middle of everything

AUTOMATIC1111:
I guess I agree with [w-e-w](https://github.com/w-e-w) here. It feels unfair to put people who put a tiny icon as an ad into the same category as people who put a gigantic banner, but an ad is an ad. If you want a donation link, you can have it on the readme page of the github.

PR #50, opened 2023-05-11
Please remove the "ads" tag.

Physton:
The following modifications have been made regarding the previously submitted PR #45:

1. The GitHub button displayed in the extension interface has been removed, and it is hoped that the ads tag can be removed as well.
2. The data storage path has been changed to within the extension directory and added to the git ignore list.
3. Translation support has been added for most languages spoken in the world. However, most of the localized language interfaces have not been done and can only be displayed in English.
4. A simple test has been conducted on all language translations for all translation interfaces, and services that are completely unusable have been removed.

Thank you for your suggestion.


w-e-w:
great, nice work
by the way we just hade some discussions about ads tag https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/49

PR #53, opened 2023-05-12
Add sd-model-preview-xd to the extension list.

n15g:
Adds model, LoRA and embedding previews to the UI including preview images, descriptions and LoRA prompt lists.

### DIFF
--- a/index.json
+++ b/index.json
@@ -1034,6 +1034,13 @@
             "description": "Improve the user experience of the prompt/negative prompt input box. It has a more intuitive and powerful input interface that offers features such as automatic translation, history records, and favorites.",
             "added": "2023-05-08",
             "tags": ["UI related", "prompting", "online"]
+        },
+        {
+            "name": "sd-model-preview-xd",
+            "url": "https://github.com/CurtisDS/sd-model-preview-xd",
+            "description": "Adds model, LoRA and embedding previews to the UI including preview images, descriptions and LoRA prompt lists.",
+            "added": "2023-05-12",
+            "tags": ["tab", "UI related"]
         }
     ]
 }


w-e-w:
found some issue
1.
don't assume model directories are there
during my testing I donlt have hypernetworks under model dir
this causes the entire extension error and to not load in
I believe this also applies for the other models diors the hypernetwork is just the first one that it encounters an errors out there
2.
dont use 	os.path.join('models','hypernetworks')] dor the default dir
see `--data-dir`
https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Command-Line-Arguments-and-Settings#all-command-line-arguments
the base directory may be changed

n15g:
The project owner is away for the weekend, but I'll see if I can fix these tomorrow and get something submitted for early in the week.

w-e-w:
the extension is working well enough
I'm going to add it to index now just remember to fix it when you can

yeah you have already opened a PR

CurtisDS:
> dont use os.path.join('models','hypernetworks')] for the default dir
see --data-dir

--data-dir is used to build out all of the paths that are in cmd_opts by the looks of it. So as long as I am using the paths from cmd_opts I dont have to manually check --data-dir right?

w-e-w:
> > dont use os.path.join('models','hypernetworks')] for the default dir
> > see --data-dir
> 
> --data-dir is used to build out all of the paths that are in cmd_opts by the looks of it. So as long as I am using the paths from cmd_opts I dont have to manually check --data-dir right?

see
https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/dev/modules/paths_internal.py
https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/dev/modules/paths.py

CurtisDS:
thanks

PR #58, opened 2023-05-15
Added Model Preset Manager

rifeWithKaiju:
https://github.com/rifeWithKaiju/model_preset_manager

Lets you create and apply presets per model for all generation data (e.g. prompt, negative prompt, cfg_scale, resolution, sampler, clip skip, steps, etc). Also automatically fetches model trigger words and thumbnails from Civitai.


w-e-w:
ISO date

PR #61, opened 2023-05-18
add Miaoshouai Assistant

triphop:
Miaoshouai Assistant allows you to enhance your automatic1111 webui user experience, including change and save boot settings, managing and downloading your models, prompts from a single tab.

### DIFF
--- a/index.json
+++ b/index.json
@@ -1083,6 +1083,13 @@
             "description": "One Button Prompt is a script for beginners and advanced users that generates an entire prompt from scratch. It is random, but controlled. You simply load up the script and press generate, and let it inspire you.",
             "added": "2023-05-14",
             "tags": ["script","prompting"]
+        },
+        {
+            "name": "Miaoshouai Assistant",
+            "url": "https://github.com/miaoshouai/miaoshouai-assistant.git",
+            "description": "Miaoshouai Assistant allows you to enhance your automatic1111 webui user experience, including updating boot settings, managing and downloading your models, prompts from a single tab.",
+            "added": "2023-05-18",
+            "tags": ["tab", "models", "UI related", "online"]
         }
     ]
 }


w-e-w:
issues
561c117351e1)
1. as the extension includes self-advertisements, as such we require the ads tag, see [prior discussion](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/49)
you could decide to tag you extension with ads (which would result in an extension be less discoverable), or remove the ads for the extension
image for details
note currently there is a bug in webui <= [1.2.1](https://github.com/AUTOMATIC1111/stable-diffusion-webui/tree/89f9faa63388756314e8a1d96cf86bf5e0663045) that will ignore `--data-dir` if it is set via `COMMANDLINE_ARGS`
this bug will be fixed in upcomming 1.3.0 and dev branch

recommendations
1. because you once included assets, your repository size is **humongous** (.git 1.11 GB)
it is possibly too late as this will result in a force push, you should use the methods documented [here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository) to clean your Repository
I intend to make a PR at webui to make it so that it only clones extension at `--depth 1` which would also resolve this issue
but until then anyone who downloads extension is massively wasting space

2. your extension downloads models to a temp directory inside the user's home `.miaoshou_assistant_download` which might not be on the same drive as the final destination of the model, this means that the model will have to be downloaded then copied
I recommend you directly download the model to the destination folder just add a .tmp suffix and rename it when the download is finished

upcoming changes to web UI
[UI selection for cross attention optimization](https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/10525) if this gets implemented then a portion of your extension might be obsolete, not totally and especially with regards to xformers

w-e-w:
wtf `civitai_models.json` why just why 485 MB json civitai catalog

w-e-w:
nope, not going to merge this
the user that that configuration is nice
but the model downloader part is essentially bloatware
@triphop @yanshoutong are you two affiliated with @miaoshouai or miaoshouai.com?
I see that yanshoutong is a contributor to the extension

either separately extension out into two different extensions or makethe downloader not a bloatware

w-e-w:
https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/10564
made a pr for depth=1
so extension doesn't take 1.2GB of space

miaoshouai:
> wtf `civitai_models.json` why just why 485 MB json civitai catalog no that's not going to work
> 
> this is not a job for git-submodules, you don't precompile a catalog and load it as assets you already online extension, just query civitai for the up-to-date information when the user needs it

Hi w-e-w, thank for the suggestions. The reason why we did the civitai-json file is because many of our users are in China and they won't have direct access to civitai unless using proxy. That is the reason why we packed the 50m gzip file as the asset and unpack it to json file during the extension installations so that their model info can be read correctly even without access to civitai.

w-e-w:
> > wtf `civitai_models.json` why just why 485 MB json civitai catalog no that's not going to work
> > this is not a job for git-submodules, you don't precompile a catalog and load it as assets you already online extension, just query civitai for the up-to-date information when the user needs it
> 
> Hi w-e-w, thank for the suggestions. The reason why we did the civitai-json file is because many of our users are in China and they won't have direct access to civitai unless using proxy. That is the reason why we packed the 50m gzip file as the asset and unpack it to json file during the extension installations so that their model info can be read correctly even without access to civitai.

@miaoshouai
okay I suppose that is a valid reason
due to containing self-advertisement I have added "ads" tag
in the future if the advertisement is removed you can make another to have the tag removed

I've made a PR to webui so that it doesn't the clone entire history of the repo
currently we are still speculating on potential issues of shallow clone
if that gets implemented it will prevent a user from having a ridiculous 1.2 GB extension

w-e-w:
it still doesn't make sense
if a use can't access civitai, then they I won't be able to download models from civitai
what is the point of having a local copy of the database

so that they can browse the model they want and then connect to a proxy then download?

miaoshouai:
this is for two reasons.
1. under model management, user of this extension can still view their already downloaded model infomation. When they click on the downloaded model, extension matches the hashs of the model and then get the model info and sample tags from the civitai json, so that they can view the tags from this model's sample covers and then send to txt2img or img2img as a starting point for their work. this will avoid them going back and force between civitai and webui to find how to use their models. we want the them be able to do that even without access to civitai.



3. there is a new prompt management tab under development, which reads the sampe tags of the model. if the user has their chatgpt api, they can simply just enter the main tag they want and the extension will instruct chatgpt to generate model specific tags based on the exsiting sample tags, trigger words from the json file, rather then general prompt might not be getting the best out of the model. (test screenshoot attached, will be able to generate negative prompt, cfg, image width and height,etc in the public version)



ps. any ads related items will be removed in the next update.

JustMaier:
@miaoshouai if we could expose access to the Civitai API via an alternative domain, do you think that would make it so you didn't have to ship the large json file?

miaoshouai:
that would be a big YES! basically all we are doing is grabbing what is already avaliable from the api and make it locally avaliable.

w-e-w:
I think it's easier to communicate if you join AUTO's discord (I assume you have a VPN of some kind)
we currently have a discussion thread on this topic
[the Discord link is on the wiki](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Contributing)

PR #65, opened 2023-05-23
Added Web UI client queue

Kryptortio:
This extension is a queue system for the web ui, it's strictly client side with everything done in the browser. Everything is mostly done by using the web ui itself to change the settings and generate images. There are [some minor](https://github.com/Kryptortio/SDAtom-WebUi-us#prompt-filter) additional features but mainly it exists to add a queue.

Initially [I made a user script](https://github.com/Kryptortio/SDAtom-WebUi-us) (main repo) for myself to save time by not having to wait for the prompts to finish before adding the next one. However since people [seemed](https://www.reddit.com/r/StableDiffusion/comments/zq0wl9/i_made_a_queue_system_for_automatic1111s_stable/) to [like it](https://www.reddit.com/r/StableDiffusion/comments/11boc0j/i_posted_a_couple_of_month_ago_about_a_queue/) I've also made [an extension version](https://github.com/Kryptortio/SDAtom-WebUi-client-queue-ext) that just contains the user script (making it easy to add fixes to both repos).

### DIFF
--- a/index.json
+++ b/index.json
@@ -1104,6 +1104,13 @@
             "description": "WebUI extension for StableSR - a powerful upscaler for both realistic and anime images",
             "added": "2023-05-22",
             "tags": ["manipulations"]
+        },
+        {
+            "name": "Web UI client queue",
+            "url": "https://github.com/Kryptortio/SDAtom-WebUi-client-queue-ext.git",
+            "description": "Adds a queue function to the web ui that processes a set of settings one or more times, it can then load a different set of settings and so on.",
+            "added": "2023-05-23",
+            "tags": ["script", "UI related", "prompting"]
         }
     ]
 }


w-e-w:
works great but I think you may not heard of the concept of dark mode
I only noticed the add to queue button after I read the docs
image

Kryptortio:
I wanted the buttons to be unobtrusive but maybe it was a bit to much, I've pushed an update increasing the opacity of the buttons.

w-e-w:
it's your call
after knowing where it is I guess it's okay but when I was trying and I spent some time finding it

w-e-w:
I am no good with JavaScript so I'm not sure if it can be done easily
maybe when you hover over it it's increases the opacity could be a good middle ground

w-e-w:
or put a opacity setting in your script settings

w-e-w:
or draw more attention to the button in your documentation
something like arrows pointing box
image

Kryptortio:
Thanks for the feedback, I did changed it to about double the opacity but the setting was a good idea so now I have added that as well.

PR #70, opened 2023-05-28
Add AWS Extension for Stable Diffusion

yike5460:
This is a WebUI extension to help user migrate existing workload (inference, train, ckpt merge etc.) from local server or standalone server to AWS Cloud.

### DIFF
--- a/index.json
+++ b/index.json
@@ -24,8 +24,8 @@
             "name": "Stable Diffusion AWS Extension",
             "url": "https://github.com/awslabs/stable-diffusion-aws-extension.git",
             "description": "Allow user to migrate existing workloads including ckpt merge, model training, model inferencing onto AWS",
-            "added": "2023-05-01",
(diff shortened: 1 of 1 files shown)


yike5460:
@w-e-w Would you please check and let me know your feedback?

w-e-w:
> @w-e-w Would you please check and let me know your feedback?

sorry for the late response

I did  check it when you initially posted it and it seems all good but I wasn't sure, but I didn't want to improve it right away because I'm not familiar with AWS, I tried asking someone else so no one respond, I'll ask others again and if there's still no response I will still approve it prove

the fact that you're documentations are partially only in Chinese doesn't help
> I can read Chinese but I don't know AWS

one thing that I would like to add is that I think this requires the `online tag`

w-e-w:
I edite the date form 05-01 to 06-01 and added the `online` tag`
no one I asked have issue with it so it's been added

PR #72, opened 2023-05-29
Add Lua scripting tab

yownas:
This PR is for an extension that adds a tab in the webui where you can write Lua-scripts to generate images.

https://github.com/yownas/sd-webui-lua


w-e-w:
lol

> This project started as a joke

nice

personally I don't understand why one would use Lua when they can just directly use python

but sure, if someone wants to try it go ahead

PR #75, opened 2023-06-05
add Model Downloader extension

Iyashinouta:


Iyashinouta:
wt f very fast response

w-e-w:
I did already check your extension yesterday
since "someone" decided to suddenly posted on the wiki
so check the extension ASAP to see if it's malicious, if it is I will remove it

Iyashinouta:
so what do you think about this extension?

w-e-w:
well it's just another downloader
and It isn't as well integrated as some and doesn't contain extra features such as metadata

someone even use it not me though some people might find this simplicity to be ince
but me personally for simple things like this I would just use the terminal or write code myself

by the way if you call this fast response what do you think about this
this is https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/3056
look at the request for review time and the merge time

Iyashinouta:
in a short time, lucky time

Iyashinouta:
By the way metadata, maybe I'll add it in the feature next time

w-e-w:
sute it's all up to you

just remember if you decide to add advertisement self-advertisement into the extension itself, you would need to add the `ads` tag, only in READM.md is ok

PR #76, opened 2023-06-05
Add style editor to the index

chrisgoringe:


w-e-w:
some issues with your extension
you should be using ` encoding="utf-8-sig"`
```py
      dts.to_csv(save_as, encoding="utf-8-sig", columns=cls.cols, index=False)
```
in webui we use utf-8 bom because this allows a stupid program called Microsoft Excel 
 to read the file without issue (if you have non English characters inside say Chinese)

don't know if it's possible to do easily, but after you write something please CSV file you  should update the database in webui
https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/shared.py#L241
otherwise there will be a desync between what's the webUI has read into memory and the actual file
furthermore if a user tried to save new sytles via the base webui button, desynced, some strange stuff would happen

most importantly
interface is implemented very dangerously, it is very easy to accidentally overwritten existing cell
accidentally press any key when selecting the cell will overwrite this selected cell
or if I try to press Ctrl+C to copy the cell will be cleared
Ctrl+C or any other key combo not working is ok
but accidentally take out the cell is not

if I'm normal user I'll I will be very angry when I see this happen
this is the most serious issue, as it would easily cause data loss to a unexpecting user
due to this don't feel comfortable putting this on index currently

I also suggest you save a backup of the file before edit (like what web UI does now)

chrisgoringe:
Thanks for feedback. I've updated the extension:

- now saves at utf-8
- updates the styles database when it saves to the master style file
- doesn't delete contents of cell on keypress (except for backspace or delete)
- supports cut/copy/paste

I've added 'create backup' to my TODO list, once I work out the best way to implement.

w-e-w:
nice~~~

PR #79, opened 2023-06-07
Update url to composable LoRA

neversay:
The original repo is broken and no longer maintained. I tested this version and it is working as expected in A1111.


w-e-w:
send notices to the respective repos
if I don't receive a negative response from them I will update the index in about 2 weeks

waltercool:
It been two weeks and original author haven't made any response. Fork author commented at https://github.com/opparco/stable-diffusion-webui-composable-lora/issues/20 and seems to be OK as replacement.

Issue #82, opened 2023-06-08
Please Add: "Inpaint Anything" 


CCpt5:
Name : Inpaint Anything

Tag : Inpaint, Segment Anything, Masking, Controlnet 

Description:  "Inpaint Anything" utilizes "Segment Anything" to automatically generate masks for use in inpainting either in its own tab, or via IMG2IMG and Controlnet via mask export.  Segment Anything enables users to specify masks by simply pointing to the desired areas, instead of manually filling them in. This can increase the efficiency and accuracy of the mask creation process, leading to potentially higher-quality inpainting results while saving time and effort.

URL : https://github.com/Uminosachi/sd-webui-inpaint-anything

----

The developer of this extension would be delighted if this extension could be added to the in app list. Thanks:

![32111111](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/assets/37972078/2cb00632-4ec4-4e40-8a5c-fd3241b2543a)

Post: https://github.com/Uminosachi/sd-webui-inpaint-anything/issues/44#issuecomment-1583430213

PR #83, opened 2023-06-09
sd-webui-inpaint-anything

w-e-w:
add sd-webui-inpaint-anything
https://github.com/Uminosachi/sd-webui-inpaint-anything
@Uminosachi
we received a request to add your extension https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/issues/82
can you confirm if the description and the tags are to your liking

### DIFF
--- a/index.json
+++ b/index.json
@@ -1188,6 +1188,13 @@
             "description": "The extension allows you to investigate embeddings/conditioning and feed conditioning to Stable Diffusion.",
             "added": "2023-06-06",
             "tags": ["script", "tab", "science"]
+        },
+	    {
+            "name": "Inpaint Anything",
+            "url": "https://github.com/Uminosachi/sd-webui-inpaint-anything.git",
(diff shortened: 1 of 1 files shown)


Uminosachi:
Please utilize the description and tags provided below. If "inpainting" cannot be applied as a tag, please use "editing".

"description": "The Inpaint Anything extension performs SD inpainting, cleaner, ControlNet inpaint, and sending a mask to img2img, using a mask selected from the segmentation output of Segment Anything."
"tags": ["tab", "inpainting"]

w-e-w:
to be honest the tag system doesn't make sense to me
currently it is used as a thing to hide extensions instead of categorizing
I hope in the future we could change it so that it's used to select certain categories of extension instead of the other way around
because of this currently I don't think it's a good idea to add it extra category
if in the future we flipped out the system worked in web UI then we could a new tag

w-e-w:
description has been updated

Uminosachi:
Could you please use the tags provided below?

"tags": ["tab", "editing", "manipulations"]

w-e-w:
I made the Tag changes

I really should have asked you to just make a PR this way you can edit it directly
if you're not okay with this fork this and make if you are and adjust to your satisfaction

PR #93, opened 2023-06-16
add BatchCheckpointPrompt

h43lb1t0:
This PR adds my Script BatchCheckpointPrompt

https://github.com/h43lb1t0/SD-WebUI-BatchCheckpointPrompt

in the script is a link to the github repo of an extension which is helpful in my script, if the link is considered advertising i will remove it

### DIFF
--- a/index.json
+++ b/index.json
@@ -21,6 +21,13 @@
     },
     "extensions": [
         {
+            "name": "Batch Checkpoints and Prompts",
+            "url": "https://github.com/h43lb1t0/SD-WebUI-BatchCheckpointPrompt",
+            "description": "Test a base prompt with different checkpoints and for the checkpoints specific prompt templates ",
+            "added": "2023-06-16",
+            "tags": ["script"]
+        },
+        {
             "name": "Kandinsky",
             "url": "https://github.com/MMqd/kandinsky-for-automatic1111",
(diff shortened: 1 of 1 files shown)


w-e-w:
did you even test your script? you don't have "scripts.settings" so it fails to import
so the entire script doesn't load, I got it to load by commenting that line out
```py
from scripts.settings import on_ui_settings
```

unintrusive links to documentation and related extensions are fine, so long it's reasonable
blatant advertising like asking for donations shoud be tag with `ads`
gray area, donation on documentation page

out of curiosity what's wrong with the built-in XYZ script? it seems to be doing basically the same thing
image

w-e-w:
also in case you aren't aware [Stable-Diffusion-Webui-Civitai-Helper](https://github.com/butaixianran/Stable-Diffusion-Webui-Civitai-Helper) has no license, which means All Right Reserved
I'm not a legal expert and I'm not sure what you used from them, but beware you may not be allowed to use their code

h43lb1t0:
I have the settings file localy, but its not in the repo yet/ I've removed the import. 
the difference to the x/y/z plot is that not only the checkpoint is changed, but also the posive prompt is inserted into a prompt specific to the checkpoint.
I'm not using any of the Civitaihelper code (i used to do). I'm just reading files created by that extension

PR #106, opened 2023-06-25
Create sd-webui-model-scanner.json

w-e-w:


w-e-w:
@ddomen
I asked what other devs think of this and this is what they said

from A:
> auto webui has a safe pickle loader that will refuse to load if the pickle does anything funny
(+ if you're remotely sane you're not download pickle files nor uploading them. safetensors should be a requirement for distribution cough civitai cough)
either way i don't think that extension adds any value over the safe unpickler
if it's approved as a "useless but why not" thing, then (run it before loading them!) needs to be removed from the description
i note that their implementation is pretty bad too https://github.com/ddomen/sd-webui-model-scanner/blob/master/scripts/sd-webui-model-scanner.py#L22-L25 they have a blacklist of naughty calls that an actual hacker could easily workaround
vs. the webui's safe unpickler having a tight whitelist which prevents that


from B:
> I think a real world attack, if there ever is one, would be a malicious update to an extension
Not a pickle

I have no objection in adding this to index, but I also agree with them that I don't think it's much use
if you still want to add it ti the index please reply

ddomen:
@w-e-w 
thank you for your reply.
If not a problem, I would like to discuss about the security of loading models, also to improve my understanding on the topic.
Before adding the extension in the index, maybe let me know further thoughts about it.

As mentioned in the extension readme, the extension is just a wrapper of other libraries used to scan the models. Bad or not, it is a start point to be aware that loading models can be dangerous.

The implementation linked by **A** comes from just one of the two scanners ([zxix](https://github.com/zxix/stable-diffusion-pickle-scanner)), the other library ([mmaitre314](https://github.com/mmaitre314/picklescan)) is actually checking the pickle raw instructions for a certain set of malevolous or suspicous actions (see code at [mmatire314/picklescan/src/scanner.py#L172](https://github.com/mmaitre314/picklescan/blob/41a70a51c1f66e4841004691bd5179ea7843bd91/src/picklescan/scanner.py#L172)).

I took a closer look to webui unpickler and it is surely **way better** than the first scanner (zxix), but I'm not sure it performs any opcode check (like mmaitre314 does). Referring to Hugging Face, there are some *"sophisticated exploit"* that allow to circumvent custom unpicklers (https://huggingface.co/docs/hub/security-pickle#potential-solutions), pretty rare on HF thanks to the use of **ClamAV**, but still an issue (and we are talking of HF, *cough civitai cough*).

Definitely I agree with **B**, but if HF cares of scanning modules I think they have their good reasons to do so. I also understand his point on "why should I trust an extension more than a model?". That's valid for any extension, that's the reason why there is this index I suppose.

w-e-w:
maybe join auto's discord server?
I am not the one to discuss about this topic
see link on webui wiki contribution

PR #112, opened 2023-07-02
feat: add qrcode-toolkit

antfu:
Wrote a blog post about the workflow: https://antfu.me/posts/ai-qrcode-refine

Thank you :)


w-e-w:
interesting workflow and integration
the only thing I don't like about this is that it's not fully local
I think during testing you self-hosted it on localhost:3000?
I assume that you make it like this because you want to make development / update easy

if you decide to make it fully local something in the future come and have the online tag remove

antfu:
Thanks! Sure, it's possible to work locally. At the current stage, I'd like to iterate faster to provide up-to-date fixes. Later once it becomes more stable, I will surely consider making it local :)

w-e-w:
to clarify this is just a suggestion I'm not forcing to make it local
is up to you on how to proceed

PR #123, opened 2023-07-11
Extension submission for Loopback Scaler

Elldreth:
This is an img2img script that I developed and would like to share in the extension index.

The Loopback Scaler is an Automatic1111 img2img script that enhances image resolution and quality using an iterative process, Somewhat like the regular Loopback script, but this includes additional features that improve on the results.


w-e-w:
I removed the invalid tag `img2img`

PR #126, opened 2023-07-11
Update index.json

limithit:
Add an extension to change the background for png images

### DIFF
--- a/index.json
+++ b/index.json
@@ -19,6 +19,22 @@
         "extras": "adds new functionality to the extras tab."
     },
     "extensions": [
+       {
+            "name": "Gandalf",
+            "url": "https://github.com/limithit/Mask2Background.git",
+            "description": "It is to give the product for the background",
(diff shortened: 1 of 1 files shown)


ClashSAN:
@limithit hi, add a new json with your extension  URL title in the extension folder.
the automation system needs you to do things that way, or things can break.

w-e-w:
in this case it wouldn't break
the entry would just be automatically removed

w-e-w:
I could fix it for you but like the description really is very unclear and do you really want to call it Gandalf

limithit:
@w-e-w Yes, I don't know much about the process. I'm sorry.

w-e-w:
> @w-e-w Yes, I don't know much about the process. I'm sorry.

well next time try reading https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions#how-to-submit-extension
I think my instruction is quite clear

PR #127, opened 2023-07-11
Add an extension

limithit:
Take a png image, generate a background and mask, then send it to img2img to generate a new background


w-e-w:
1st that json is not valid
2nd you clearly didn't read the instructions, you don't need to add the added date
3rd only existing tags are allowed, I removed it
I fix it for you

PR #128, opened 2023-07-12
updated to v1.6, added using gpt for prompting, vram releasing, all m…

miaoshouai:
updated to v1.6, added using gpt for prompting, vram releasing, all model cover downloading, restructed assets to reduce extension size, removed ads

### DIFF
--- a/extensions/miaoshouai-assistant.json
+++ b/extensions/miaoshouai-assistant.json
@@ -1,13 +1,12 @@
 {
     "name": "Miaoshouai Assistant",
     "url": "https://github.com/miaoshouai/miaoshouai-assistant.git",
-    "description": "Miaoshouai Assistant allows you to enhance your automatic1111 webui user experience, including updating boot settings, managing and downloading your models, prompts from a single tab.",
+    "description": "Miaoshouai Assistant allows you to enhance your automatic1111 webui user experience, including updating boot settings, freeup vram, managing and downloading your models from civitai.com and liandange.com, generate prompts from chatgpt all through a single tab.",
     "added": "2023-05-18",
     "tags": [
         "tab",
         "models",
         "UI related",
-        "online",
-        "ads"
+        "online"
     ]
 }


w-e-w:
currently asking for opinions

before https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/61
image
now
image

ClashSAN:
@w-e-w Infinite zoom extension also has those. It should be ok.

On this subject, the Lobe theme extension has "More Products" in the ui, which are all free opensource github projects. I don't know if the "ads" tag should mean paid products only.

Maybe if there is a place in the settings panel where extensions can post support info and advertise. Then it would be noninvasive for the user and no need for "ads"

w-e-w:
my personal opinion is that donation links are ads links to other things that are intrusive are also considered ads

for this one it's in ok for me, but I want to asking for opinions

> Maybe if there is a place in the settings panel where extensions can post support info and advertise. Then it would be noninvasive for the user and no need for "ads"

there was a prior discussion on whether or not to add a section for donation links
this is if we do that then the question is should a section for the webUI itself
and then it will be a question of who gets to put their link on that page, it will be an entire mess and I don't think no one wants to deal with
> I think the conversation on Discord

miaoshouai:
@w-e-w we are still shown as ads in the index.json ><

w-e-w:
I know, as I haven't merge it, was jsut fixing the github actions automation and merge conflicts

I'm not sure how we going to proceed
maybe we will adjust the rules, maybe we'll create a section dedicated for ads, maybe we will 
I'm not sure

miaoshouai:
> I know, as I haven't merge it, was jsut fixing the github actions automation and merge conflicts
> 
> I'm not sure how we going to proceed maybe we will adjust the rules, maybe we'll create a section dedicated for ads, maybe we will I'm not sure

ok~ i think for our case, we only have discord and QQ group there so that people can join and give us feedbacks for the extensions. the new version should be pretty straight forward. It contains no ads.

w-e-w:
I am still unsure how we will proceed the future
but considering that your extension is not the biggest offender
let's remove the ads tag

miaoshouai:
> I am still unsure how we will proceed the future but considering that your extension is not the biggest offender let's remove the ads tag

OK~ thanks w-e-w~ it's showing up in the list~

PR #135, opened 2023-07-15
tag violation Nevysha's-Cozy-Nest

w-e-w:
@Nevysha missing `online` tag

### DIFF
--- a/extensions/Cozy-Nest.json
+++ b/extensions/Cozy-Nest.json
@@ -4,6 +4,7 @@
     "description": "Extension for an improved UI/UX. Find your cozy spot on Auto1111's webui!.",
     "added": "2023-05-02",
     "tags": [
-        "UI related"
+        "UI related",
+        "online"
     ]
 }


Nevysha:
Which part of Cozy Nest strongly require online feature ? Cozy Nest rely on internet to fetch latest version and patchnote, which is true for every other extension.

w-e-w:
> online tag is Required for any extension that connections to external server during regular use aside from one time downloading of assets.

I have to admit your extension kind of got caught in the crossfire of [lobe theme](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/136), normally I would let it slide
indeed your extension doesn't strongly requires internet connection to function
but as you said it does has auto update checking
image
this part of the UI doesn't work without a connection, one can make the argument that a integral part of your UI is not working, (I will consent to that this is the weak argument)
image

webui extension updates are opt-in, user will have to trigger it or manually enable using `--update-all-extensions` `--update-check`
extensions generally don't have a built in auto update / update checker
your Cozy Nest automatically checks for updates and changelog of your own extension by **default**
only good thing to say about is that the server that is access is GitHub, which I in my knowledge don't expose access count IP to repo owner
whether or not if GitHub themselves is using his data internally is unknown

webui itself also have some issues that not everything is totally local
and some people really don't like these other things examples
https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/11378
https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/11751
personally I think these two use cases are perfectly acceptable

one time downloading of assets such as models don't count as online

if you wish to dispute `online` tag, I'm open to conversation

>by the way
as we started to collect metadata for extensions, it is possible that we can Implement a opt-in feature that can automatically check for updates without fetching from individual repo

Nevysha:
>only good thing to say about is that the server that is access is GitHub, which I in my knowledge don't expose access count IP to repo owner

Yes that's exactly why I'm relying directly on the repo to fetch version and patchnote. Althought it's "by default" it works totally offline. And sure the patchnote will not display when offline, which is intended.

Cozy Nest aimed to add UX improvment to the native gradio UI, and seeing if something, such important as the frontend, is up to date or not with a simple view is an improvment. Yet it does not check on a unknown server. Again, the goal of Cozy Nest is to make a1111 more accessible to non tech heavy user, I won't make it as an option to check or not if there is update.

I does not agree with the way you just changed the tags neither why you did it. I'm not prone to deeply argument more than what I write above so either the online tag is removed, or remove Cozy Nest from the """""official"""" extension list.

Regards

Nevysha:
Moreover, the way extensions works in a1111 allow every developper to just push malicious code in the preload.py (like delete all file/folder on the computer, or encrypt them). So I think this discussion is ironic.

w-e-w:
in case we're not on the same page
the `online` tag **does not** mean the extension will be hidden by default
> this is unlike the `ads` which will be hidden

`online` it's merely **a notification to users** that a certain extensions contains elements that request to an external server regularly when operation

and I'm not requesting any change from you
if you wish to continue to fetch update data I am not against it

with this use case considering that the server is GitHub I think it's fine to not use `online` tag in this case
if the server is a non well known server then there will be no argument that it requires `online` tag

do you still request the tag to be removed?

sasha-jan:
yes please, because the core features of Cozy Nest does not need to be online to works.

Nevysha:
yes please, because the core features of Cozy Nest does not need to be online to works.

w-e-w:
done https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/139

w-e-w:
See if your viewpoint correctly
> "You belive your extension does not require Internet access to function thus categorizing it as `online` is a miscategorization"

While other some belive the lack of `online` meant that it should be totally "local"

Because we don't have a clear distinct category I have to uphold it with a higher standard (then I would like)

Do you think creating a new category like `local` would help with the situation
This would be exclusively used for extensions that absolutely does not access the Internet for any reason after an initial installation

My current plan is
`local` No Internet access of any kind after initial install
`online` Requires Internet to function
If these tags are not used: Extension might access Internet for certain operations but it is not a core part of the extension itself

I'm not sure how would we categorize one time download of assets after the initial

and yes, some people do care about one time download
https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/11518
>personally I think these people are a bit unreasonable

w-e-w:
since you did not reply to this I can ony assume that you don't have a opinion or has just missed the comment
current planning going forward with the `online` and `local` tags
with non-specified being gray area for extensions that would have some limited internet access time to time

Nevysha:
Sorry, have a lots of things to do lately.

Why not a "limited" for non required internet aside from some non core operation such as update with a white list of endpoint ? (Huggingface, github, civitai,...)

w-e-w:
thie issue some people concerns is not just about security
for people who live in China they can not access some reasorces

in fact all 3 you just listed are blocked in some capacity
civitai I think to completed block
github see to work in som region and not in other (but there is proxy / mirrors of gihub) 
Huggingface for what I heard is sometimes block??? idk since I don't live in china
they even have trubble installing Torch !!! or using pip without extra configerations

TLDR is that, some people would require it to work without access  to servers because may not work

and I agree with them to some degree
if a software is `local`, ment that the software can theoretically work "forever"
while something that need external server ment that it might break sometime in ther futhere


another method of approch is that we can list what servers thet a extension will access

this may tie into anothe topic
some extensions need to have higher execution priorectory before other extensions 
there is an idea about havening some meta data to denote that
if we do have this metedata in a file, then it may be possible for extensions  to also declear more info about the extensions

everything for a changlogs to noitces to user before install, and theis can be change for the extension side and not hear in the index
it can also have a list of servers that it access
so the user can decide them selfs with mo detal info

IMO the tag system is not good
webui has become to large for this
but it is what we have to work with for now
if possible I would like to replace it with a better system

> with the ahugelimit backwards compatibl 😭

Nevysha:
Maybe the extension json might include a link to a file on the main repo for a list of requirement and display that link in the ui ?

Nevysha:
Also the difference between an extension that need access to internet and another that may for metadata fetching shoud be differencied as we said above

w-e-w:
> Maybe the extension json might include a link to a file on the main repo for a list of requirement and display that link in the ui ?

that is kind of the idae, a file that defines stuff about the extensions including

extensions version to webui version matching (webui version to extensions commit hash)
this can be use to install diffrent version as aposes to the newest version if nessery (like a newversion of an ext only support newversion of webui, the the autohr can say "if webui 1.4 use commit XXXX" etc)

priorectory of each call back

internet requirements
onetime download for xxx 
get meta data for yyy
requires connections to zzz
none after install frome git pip
etc

or other info like
this ext only works on nvdia gpu
no longer maintained, use switch to fork
thers alot of thisng that can be usefue

the idea is to have this repo auto grab those data and compile it into the index so it can be fetch all at once bby webui and not have to fetch from each individual repo

PR #142, opened 2023-07-18
sd-history-slider

LucasMali:
A slider change the prompt based on the history of what was used.


w-e-w:
tip, configer git to ignore `__pycache__`

PR #145, opened 2023-07-22
sd-dynamic-javascript extension submission

pmcculler:
Followed the instructions, I think!
I believe more than a few people will find this useful, particularly professional and hobbyist programmers. Examples of use are in the extension readme; more complex examples to really show off what you can do coming, but already what's there should let imaginative people do fairly amazing things, IMO.


w-e-w:
interesting
by the way do you know there is a `--allow-code` arg that gives you a "custom python scripting box"
image

w-e-w:
you have extra comma at the end of the last tag

PR #149, opened 2023-07-23
Create Simplified_Chinese_Language_Pack

StormyOrange:
Simplified Chinese translation of all WebUI texts since v1.0.0-pre, and corrected unreasonable translations by dtlnor. Also translated a large number of extensions by other authors into simplified Chinese, with a total of 6000+ words translated.

### DIFF
--- a/extensions/Simplified_Chinese_Language_Pack.json
+++ b/extensions/Simplified_Chinese_Language_Pack.json
@@ -1,0 +1,9 @@
+{
+    "name": "Simplified Chinese Language Pack",
+    "url": "https://github.com/StormyOrange/SD-WebUI-Localization_ZH_CN.git",
+    "description": "Display WebUI interface in Simplified Chinese",
+    "tags": [
+        "UI related"
+    ],
+    "added": "2023-07-24"
+}


w-e-w:
@StormyOrange 
you miss the `.json`

@dtlnor thought you might want to know about this

dtlnor:
@StormyOrange @w-e-w 
1. There should have only one translation for each locale language suggested by Automatic1111, a better idea is pull request zh_CN repo. NOT creating a new version based on the originial zh_CN without forking, without including the MIT license with repo owner's name which is me. And no reference at all.
2. Most of the translation are revised by multiple contributer. Some of them are compromise within multiple version. Thats unreasonable to say "corrected unreasonable translations" without giving any description.
3. The extension tag should be "localization" not "UI related"
4. Please keep the naming style same (at least similar) for the localization file, not `简体中文语言包_StormyOrange.json`

w-e-w:
I'll let you handle this

> The extension tag should be "localization" not "UI related"

I miss that one there was two PR and one has localization the other is not but the other is editing the index directly

btw, you don't need to edit the index directly to remove, if you remove it from the extensions directory it will automatically be removed from index by the workflow

PR #171, opened 2023-08-15
add sd-webui-models-manager

dhs964057117:
## Checklist:

### DIFF
--- a/extensions/sd-webui-models-manager.json
+++ b/extensions/sd-webui-models-manager.json
@@ -1,0 +1,9 @@
+{
+    "name": "sd-webui-models-manager",
+    "url": "https://github.com/dhs964057117/sd-webui-models-manager.git",
+    "description": "An extension that allows users to customize model file paths",
+    "added": "2023-8-15",
+    "tags": [
+        "tab"
+    ]
+}


w-e-w:
1. you fail to communicate what your extension does, I have to read your code to understand the functionality
2. your description is incorrect, you're not "can change your models path" you're creating symbolic links
3. the average user won't understand what is a symbolic link or Junction or and Hard links
4. normally on Windows you cannot create symbolic links unless you have special permissions, typically you either have to have Administration rights or (I've heard) enabling "Developer Mode" also works, you can also configure it in group policy editor but that's not available for all users, used Junctions for directories and hard links for files if you must
5. why not just put the entire Target directory under a junction inside the current model directory why create the first level?
6. I suspect this is the first time you ever worked with gradio, because the way you implemented things is "very strange" to say least, using lots of unnecessary global variables to pass variables around, see gradio docs for a basic example on how to properly code your UI and function call https://www.gradio.app/docs/button

why do people need this, a user can set their model directory path?

dhs964057117:
> 1. you fail to communicate what your extension does, I have to read your code to understand the functionality
> 2. your description is incorrect, you're not "can change your models path" you're creating symbolic links
> 3. the average user won't understand what is a symbolic link or Junction or and Hard links
> 4. normally on Windows you cannot create symbolic links unless you have special permissions, typically you either have to have Administration rights or (I've heard) enabling "Developer Mode" also works, you can also configure it in group policy editor but that's not available for all users, used Junctions for directories and hard links for files if you must
> 5. why not just put the entire Target directory under a junction inside the current model directory why create the first level?
> 6. I suspect this is the first time you ever worked with gradio, because the way you implemented things is "very strange" to say least, using lots of unnecessary global variables to pass variables around, see gradio docs for a basic example on how to properly code your UI and function call https://www.gradio.app/docs/button
> 
> why do people need this, a user can set their model directory path?

hello，you are right, I am a novice developer of Gradio
The use of symbolic link is only to save space on the current SD installation hard drive, so that users can move model files to other hard drives. As for the permission issue you mentioned, I will explain that you need to use administrator privileges to run webui.bat

And I found that using symbolic link is the method mentioned by webui author automatic in his document. Perhaps you can say that my code style is not standardized enough, but I think there should be no problem with this implementation approach

image

w-e-w:
you should Never Never run web UI with Administration privileges

on Windows or at least NTFS you should create Junctions or hard links if you want to do so to achieve the same effect without elevated permissions

dhs964057117:
> you should Never Never run web UI with Administration privileges
> 
> on Windows or at least NTFS you should create Junctions or hard links if you want to do so to achieve the same effect without elevated permissions

I understand what you mean, but why not run webui with administrator privileges?. I will refer to your suggestions for improvement. Thank you

dhs964057117:
> > Perhaps you can say that my code style is not standardized enough, but I think there should be no problem with this implementation approach
> 
> I don't care about your coding style I'm just suggesting that you might want to learn a better way that's all
> 
> all I care is that the user understands what will be done to their files by your extension, and that it will work
> 
> I understand what you mean, but why not run webui with administrator privileges?. I will refer to your suggestions for improvement. Thank you
> 
> webui is literally a web server that has the ability to inject arbitrary python code that will be executed blindly by your operating system it is a security nightmare
> 
> I myself don't dare open my webui instantly open internet

Oh, I see. Actually, I'm just an amateur and I don't have a comprehensive understanding. Thank you for answering me. I will find ways to improve. If there is no way, I will tell the user how to create links for each model separately without using administrator privileges

w-e-w:
as mentioned for Windows NTFS you can use Junctions and Hard links and not symbolic links, thes are effectively the same for your use case
I would know because I created an application for myself that extensively uses hard links and Junctions https://github.com/w-e-w/Wallpaper-Engine-Hard-Link-Backup
note Junctions are only for NTFS file system, it doesn't work on others like  FAT32

on Unix you should be able to use symbolic link without special privileges

regardless of the method you choose you would likely need a method to detect what OS and what file system the files resides on 
to finish it's okay if you're not able to make it work on multiple systems
as long as your extension communicates to the user properly on what it does and what it doesn't do
if you want you can explain your extension only works under certain conditions

if this extension can only work under admin privileges then I will NOT approve it to be on the index as it is will encourage people to run webui with admin privileges

because the nature of your extension it is imperative that you explain what your extension does and what the implications are
> such as if misused it has the potential to serve your entire computer onto the internet

I'm closing this PR, open a new PR if you have progress on your extension

not to discourage you but my personal opinion is that education and communication is the hard part
and if someone is able to understand what are Symbolic links Hard links Junctions, then these people would probably be able to run the commands themselves so they don't need your extension
on the other hand if someone doesn't understand what these are then they should also probably don't use extension as they probably would cause "harm"

dhs964057117:
> as mentioned for Windows NTFS you can use Junctions and Hard links and not symbolic links, thes are effectively the same for your use case I would know because I created an application for myself that extensively uses hard links and Junctions https://github.com/w-e-w/Wallpaper-Engine-Hard-Link-Backup note Junctions are only for NTFS file system, it doesn't work on others like FAT32
> 
> on Unix you should be able to use symbolic link without special privileges
> 
> regardless of the method you choose you would likely need a method to detect what OS and what file system the files resides on to finish it's okay if you're not able to make it work on multiple systems as long as your extension communicates to the user properly on what it does and what it doesn't do if you want you can explain your extension only works under certain conditions
> 
> if this extension can only work under admin privileges then I will NOT approve it to be on the index as it is will encourage people to run webui with admin privileges
> 
> because the nature of your extension it is imperative that you explain what your extension does and what the implications are
> 
> > such as if misused it has the potential to serve your entire computer onto the internet
> 
> I'm closing this PR, open a new PR if you have progress on your extension
> 
> not to discourage you but my personal opinion is that education and communication is the hard part and if someone is able to understand what are Symbolic links Hard links Junctions, then these people would probably be able to run the commands themselves so they don't need your extension on the other hand if someone doesn't understand what these are then they should also probably don't use extension as they probably would cause "harm"

ok，you are a great person, and I am not discouraged. I am glad that you can tell me where my problem lies and provide your suggestions. I will refer to your suggestions and make improvements to the program. Thank you very much

PR #178, opened 2023-08-22
add add sd-webui-models-manager.json

dhs964057117:
## Checklist:

### DIFF
--- a/extensions/sd-webui-models-manager.json
+++ b/extensions/sd-webui-models-manager.json
@@ -1,8 +1,8 @@
 {
     "name": "sd-webui-models-manager",
     "url": "https://github.com/dhs964057117/sd-webui-models-manager.git",
-    "description": "An extension that allows users to customize model file paths",
-    "added": "2023-8-15",
+    "description": "An extension that allows users to change model folder location",
+    "added": "2023-8-22",
     "tags": [
         "tab"
     ]
--- a/extensions/sd-webui-models-manager.json
+++ b/extensions/sd-webui-models-manager.json
@@ -1,0 +1,9 @@
+{
+    "name": "sd-webui-models-manager",
+    "url": "https://github.com/dhs964057117/sd-webui-models-manager.git",
+    "description": "An extension that allows users to customize model file paths",
+    "added": "2023-8-15",
+    "tags": [
+        "tab"
+    ]
+}


w-e-w:
great you make your extension Windows compatible without admin privileges

but I hate to tell you this, your extension will only work on the base default configuration of web UI

first you just assuming that people put their models in the models directory which is not the case
lots of people specified their directory using `--XXX-dir` args
like me I have set a specific lora directory, but your extension would assume that my lora are in the default dir and creat links there, so the links will all be in the wrong folder and will not work
(there are some cases that webview I would also checks for the default directory, like for checkpoints, so your extension will work in this case but for other things it won't)
and as your extension cross through all the directories and create input fields for all them you also need to actually get the real directory for those extension too otherwise you'll still change this misdating

instead of assuming that every folder inside the model directory is a directory, that user should put their model in you should get the actual directory where the models are actually being read from, unfortunately there isn't a uniform method


several things to take note
if you actually want to make this work
`--data-dir` is able to move the entire user data directory including models
args like `--ckpt-dir` `--vae-dir` `--lora-dir` is able to change the directorys, and unfortunately there are many of these args scatter webui
in some cases like `--lora-dir` will only exist if you have the lora extension enabled, most people will have to enabled as it can be but you need to account for the exception if someone doesn't enable it

since you crawl through the entire directory you're including Target directories for other extensions as well such as `controlnet` `adetailer`
truth be told I doubt you can account for what extensions does so I suggest you limit your extension to be only create links for the important ones such as checkpoints in lora (like before) otherwise it's lots of work

another issue your extension have is that you're creating json file, from what I can read from the code that file is totally unnecessary, writing to file will not necessary is unreliable to say the least but I won't really complain about this as this is your extension
but the issue is that you include this file with your extension which means that it is tracked by git
meaning that they will have a hard time updating extension, as last file has changed, if it must have a json file, delete it and added to gitignore, have your extension Auto generate the file so that it's not part of your repo

since we're talking about unnecessary files your extension contains files that I've generated from python that should not be included in your repo https://github.com/dhs964057117/sd-webui-models-manager/tree/master/scripts/__pycache__

one more thing I would highly suggest that you change your extension name to something else because you don't really manage your models

your extensions only links maybe you will call it import models from other directories
I would change your repository into something like `model importer` `model linker`

and I will change the description to something like
import models from external directories using links file

to be honest, as opposed to using extension to basically patch this function into web UI
I would suggest that you dropped this project and then make a PR to web UI to allow user to specify multiple directories for models
this in my opinion is the better option that will allow a user to store multiple directories of models across different drives
if I have time I might actually try and make this come true in web UI

dhs964057117:
ok,i kown your means,Thank you for your advice,I will modify the few shortcomings you mentioned. Can my extension be added to the extension index?

w-e-w:
> ok,i kown your means,Thank you for your advice,I will modify the few shortcomings you mentioned. Can my extension be added to the extension index?

Yes, it can be added to index
the reason why I haven't added it into index is that
1. your extension has potential bugs (if the issue isn't big then it can still be on the index but I would prefer that those bugs are not there)

> currently your extensions won't cause people's computer to break, but it doesn't work completely, it barely passes the minimum bar, even limit your extensions ability so that it doesn't function when it shouldn't or make it perfect

2. your description does not match the functionality of your extension.
> from you readme
>> can change your models save path.

>that is not what your extension does
you haven't communicate how your accenture achieve the functionality
and haven't communicate that one should disable the extension security reasons when the dont need it
you also haven't communicate that how one can remove those link files once created
(yes computer savvy uses like use for "likely" know how to do so, but since you failed communicate, that you are actually creating link files and not moving directors even people like us can get confused won't know unless they look at you code)

the reason I do reviews before adding extensions is that because webui has a huge user base any issues will be magnified
even though we don't have responsibility of what extension does, but I still feel responsible to do filtering before adding things to index

I will reiterate again
your extension can be added to index but it has to
1. have no obvious bugs
2. and more importantly communicate what it does clearly to the user

w-e-w:
make a new PR when there is progress

PR #180, opened 2023-08-24
CivitAI Browser add-on

BlafKing:
## Info 
I've submitted an add-on which let's you browse the content of CivitAI, it lets you download, update and delete any models.
Shows which models are installed or outdated and also has the ability to save associated trained tags of any model.

This is a fork which I'm currently maintaining by adding features/cleaning up the UI.

https://github.com/BlafKing/sd-civitai-browser

## Checklist:

### DIFF
--- a/extensions/sd-civitai-browser-plus.json
+++ b/extensions/sd-civitai-browser-plus.json
@@ -1,0 +1,10 @@
+{
+    "name": "CivitAI Browser Plus",
+    "url": "https://github.com/BlafKing/sd-civitai-browser-plus.git",
+    "description": "This extension allows you to browse and download any model from CivitAi without leaving WebUI!",
+    "tags": [
+        "script",
+        "tab",
+        "online"
+    ]
+}


w-e-w:
I took a look at the extension and it seems all good
normally I would add it to extension index but I have noticed that you is a fork of another repo
this causes some issues
normally we prefer to add only the root repo, if I root repo is no longer maintained then we might switch to a fork 
I have noticed that the original repo hasn't been active for a while but I have not tested if it's still working or not

so like you to answer
1. do you know what upstream status is, have you tried collaborating with them
2. or this is substantially different from the original repo and is a new extension, you have no intention to merge with the Upstream repo

if you have no intention to merging with the upstream repo then it's better for you to use a different name
one reason we do this is that currently there will be a name collision issue, as only one extension of the same name can be installed at the same time, 
> I do wish to improve the extension system so that we are able to accept forks without issue reports but currently it is not the case

TL;DR
if you have no intention of merging then the the upscream repo, you need to have a different repo name from the original
otherwise we prefer you to merge with the upstream repo and we will add the upstream to the index

note you don't need to worry that changing names will break existing installations as GitHub will automatically read directly all URL to the new one

BlafKing:
Thanks for taking the time to review the extension!
I've tried contacting the original repo to no avail, and by now there's many different features to the point that it's pretty much a different add-on.

I've gone ahead and changed the repo name!

PR #189, opened 2023-09-09
PromptsBrowser extension application

AlpacaInTheNight:
## Info 

https://github.com/AlpacaInTheNight/PromptsBrowser.git
An extension to quickly work with, explore, and catalog prompts. SD restart required after installation.

## Checklist:


w-e-w:
issues
1. make work with `--data-dir`
2. don't assume the extension path there's no guarantee that your extension will be installed under `extensions/PromptsBrowser/` the directory name is not guaranteed
3. don't put stuff in webui root, I recommend you put stuff in your extension's directory, remeber to add them to `.gitignore`
4. I don't see why you would need to start a different API server under different process, if I'm understanding correctly your code will not be able to work if people are accessing remotely, like when using gradio `--share` or `--listen`, some examples of extensions that adds their own API endpoints https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/13103#discussioncomment-6925574 using the correct method of creating API if done correctly it wll work over `--share` and `--listen`

tips
I'm not sure but it seems like you're new to python
```py
import os
print(os.path.abspath(__file__))  # current script abs path
print(os.path.join("you", "path", "to", "whatever"))  # no need to use + os.sep +
```
also remeber to remove files that don't belong such as `.vscode`

AlpacaInTheNight:
Thank you for your review. Yes, I'm a new to Python, so I don't know a lot about it and its ecosystem. I will study examples you provided and try to rewrite extension to comply with this requirements.

AlpacaInTheNight:
I made changes to decouple the extension from the local server and installation directory. The extension now works on FastAPI. I also tested it with a non-standard working folder installed via --data-dir, renaming the directory with the extension, non-standard port. I have not tested online hosting of the application, but I believe the extension should work in this mode. I also added a read-only mode just in case.

w-e-w:
I did some tests and it does seems to work with `--share`

PR #195, opened 2023-09-24
added entry of extension

sixf0ur:
## Info 
[](https://github.com/sixf0ur/sd-webui-notify.git)

## Checklist:

### DIFF
--- a/extensions/sd-webui-notify.json
+++ b/extensions/sd-webui-notify.json
@@ -1,0 +1,11 @@
+{
+    "name": "sd-webui-notify",
+    "url": "https://github.com/sixf0ur/sd-webui-notify.git",
+    "description": "Receive a host system notification when the generation is finished.",
+    "tags": [
+        "script",
+        "UI related",
+        "query",
+        "extras"
+    ]
+}


w-e-w:
Major issue

> you can define a CMD command that should be executed on the host machine

you realize that is a security nightmare right?
I have a working progress extension that also allows users to define a custom command
but I put in tons of security blocks to prevent the command from being changed when webui is remote accessible 
https://github.com/w-e-w/sd-webui-custom-autolaunch#security-warning--why-command-is-not-editable

I suggest you either implement something similar to mine or let users manually define a text file that will be executed that is not accessible from the UI
or remove the function entirely

---

is there a reason why you don't make a wrapper function to inject to your code after `shared_state` `State` `end()`

https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/5ef669de080814067961f28357256e8fe27544f4/modules/shared_state.py#L122-L128

---

btw are you aware webui has built-in functionality that when it's in browser mode it can send a notification on job complete
`Settings > Actions > Request browser notifications`

sixf0ur:
Thank you for your feedback and for pointing out the built-in functionality of web UI for sending notifications on job completion. I must confess, I didn't notice the existing notification feature in the web UI, and your observation has made it clear that there's no need for the extension to handle notifications separately.

PR #198, opened 2023-09-29
Added Ranbooru

Inzaniak:
## Info 
[https://github.com/Inzaniak/sd-webui-ranbooru](https://github.com/Inzaniak/sd-webui-ranbooru)
Ranbooru is an extension to get prompts from different boorus using their APIs. It also has additional features to dinamically change the prompt.

## Checklist:

### DIFF
--- a/extensions/sd-webui-ranbooru.json
+++ b/extensions/sd-webui-ranbooru.json
@@ -1,0 +1,11 @@
+{
+    "name": "Ranbooru",
+    "url": "https://github.com/Inzaniak/sd-webui-ranbooru.git",
+    "description": "Get random prompts from different boorus with a lot of creative features to influence the prompts. Works with txt2img and img2img",
+    "tags": [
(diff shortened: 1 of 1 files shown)


w-e-w:
why is this implemented as a script as opposed to a always on extension?

Inzaniak:
Hi! Currently, this code is implemented as a script because when I started developing it, it had a much smaller scope than now. I was thinking of converting it into an extension, but I didn't have time to take a look at the documentation to understand how to proceed. It is definitely something that I am considering.

w-e-w:
understood
I'll add this extension into the index now and you can do your development as you like

I don't think it would be that hard it is possible that it's very similar to what I did with the other extension
see PR https://github.com/pk5ls20/sd-webui-controlnet-fastload/pull/1

sadIy don't have time now to help you converted to the normal extension like I did with the other extension

PR #200, opened 2023-10-01
add nsfw_barcensor.json

MrArsenic621:
Hi 
This extension is for censoring nsfw parts of the image genrated by sd-webui
I use "stable-diffusion-webui-nsfw-censor" as a base which i used, but that repo just censor the whole thing and there was no clue what is generated so i embed nudenet into it and used it to detect and barcensor nsfw part witch currently are "EXPOSED_ANUS",     "EXPOSED_BREAST_F",  "EXPOSED_GENITALIA_F",  "EXPOSED_GENITALIA_M" 
what i created has a long way to develop, maybe be able to select whitch part to censor and don't
and indexing it will make people using it whitch is helpful to them and also might make people try to make it better and results in a good nsfw content moderator

Thank you.
have a nice day.

### DIFF
--- a/extensions/nsfw_barcensor.json
+++ b/extensions/nsfw_barcensor.json
@@ -1,0 +1,8 @@
+{
+    "name": "NSFW-Barcensor",
+    "url": "https://github.com/MrArsenic621/nsfw-barcensor.git",
+    "description": "Detect nudity and non-worksafe images using Nudenet and use barcensor to cover sensitive parts.",
+    "tags": [
+        "script"
+    ]
+}


w-e-w:
are you sure that your extension works because I couldn't get it working?

w-e-w:
I fix the `img_path = f"./extensions/nsfw-barcensor/scripts/tmp/{i}.png"` part but
```
    AttributeError: 'NudeDetector' object has no attribute 'censor'
```
I don't see NudeDetector has a `censor` only `detect`

PR #205, opened 2023-10-12
added TensorRT

contentis:
## Info 
Adding TensorRT acceleration for UNet.

### DIFF
--- a/extensions/TensorRT.json
+++ b/extensions/TensorRT.json
@@ -1,0 +1,9 @@
+{
+    "name": "TensorRT",
+    "url": "https://github.com/NVIDIA/Stable-Diffusion-WebUI-TensorRT",
+    "description": "TensorRT acceleration for Stable Diffusion on NVIDIA RTX GPUs.",
+    "tags": [
+        "tab",
+        "models"
+    ]
+}


w-e-w:
@contentis ahhh...... I think the repository is set to private

w-e-w:
there are some issues

1. One I have made a draft [PR](https://github.com/NVIDIA/Stable-Diffusion-WebUI-TensorRT/pull/1) about, whitch can cause the TRT UI to not load

2. The use of shared.opt in installer, not only that share is a very heavy import whitch can slow down the load time of webui, but more imortently, it may not be able to get the correct config file so won't really work<br>
you shoud be able to add sd_unet to quick settings at `before_ui_callback`
I will make a PR for this later<br>
> IMO, it can be good to auto add sd_unet to quick settings, but user should have the choice to remove it they should don;t wan't it

3. the repo name is `stable-diffusion-webui-tensorrt` same as AUTOMATIC1111's https://github.com/AUTOMATIC1111/stable-diffusion-webui-tensorrt, curently how it works it that it will try to install into the folder for the repo name by default, if someone has already installed AUTOMATIC1111's stable-diffusion-webui-tensorrt, it will cause installation to fail
the best solution is for use to implement something in webui to deal with this issue, I did have a proposal of appending the author owner before the repo name, but there may be some issue with this method so it was not implement<br>
the easy solution is for you to change the repo name, but I guess there are rules at NVIDIA so not an options<br>
for now I think the best solution is for AUTOMATIC1111 to his TRT repo name to prevent future issues, but people with auto TRT installed will still have issues

4. system default encoding issue, fix [PR](https://github.com/NVIDIA/Stable-Diffusion-WebUI-TensorRT/pull/2)

contentis:
@w-e-w Just wanted to let you know that the repo is public now, but apparently, you were faster :D Thank you for your input and the PRs; this is awesome! 


> IMO, it can be good to auto-add sd_unet to quick settings, but user should have the choice to remove it they should don't want it

Agreed, that's why I tried to set it in the installer, so it is a one-time call. If `shared` takes a long to load, I could try to read the config.json and just write it there, but that seemed a bit hacky.

w-e-w:
> Agreed, that's why I tried to set it in the installer, so it is a one-time call.

well, installer is not called one time, one it also called on every restart by the launch.py
as I sad, I belive `before_ui_callback` should work, I see what I can do later when I have time
got to go now

and by the way, I still have trubble getting TRT to work on my PC earlier today, not sure why yet.

w-e-w:
> but to modify the user’s custom settings is unnecessarily intrusive.

I do agree in part, from my viewpoint it if we just a line of text in a description saying "tip: recommended to add `sd_unet` to Quick Settings for easy access", but the issue is people don't seem to read, so I don't know

w-e-w:
potential issues
the names given to the `.trt`s is very long I fear it might approach Windows limit
image

w-e-w:
one major issue that I'm being experiencing randomly is this
it almost appear every time I try to export an engine
sometimes when this happens the process completely holds other times it continues I managed to get one successfully completed and use it to run sd with trt and I can say that it is significantly faster
but most of the time my engine fails to export, making this almost unusable
```shell
Building TensorRT engine for B:\GitHub\stable-diffusion-webui\models\Unet-onnx\Anime_Anything-V3.0_Anything-V3.0-fp16_Anything-V3.0-pruned-fp16_38c1ebe3.onnx
: B:\GitHub\stable-diffusion-webui\models\Unet-trt\Anime_Anything-V3.0_Anything-V3.0-fp16_Anything-V3.0-pruned-fp16_38c1ebe3_cc86_sample=1x4x64x64+2x4x64x64+
8x4x96x96-timesteps=1+2+8-encoder_hidden_states=1x77x768+2x77x768+8x154x768.trt
ERROR:asyncio:Exception in callback H11Protocol.timeout_keep_alive_handler()
handle: <TimerHandle when=355099.875 H11Protocol.timeout_keep_alive_handler()>
  File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\h11\_state.py", line 249, in _fire_event_triggered_transitions
    new_state = EVENT_TRIGGERED_TRANSITIONS[role][state][event_type]
KeyError: <class 'h11._events.ConnectionClosed'>

During handling of the above exception, another exception occurred:

  File "C:\Programs\Python\3.10.6\lib\asyncio\events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 363, in timeout_keep_alive_handler
    self.conn.send(event)
  File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\h11\_connection.py", line 468, in send
    data_list = self.send_with_data_passthrough(event)
  File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\h11\_connection.py", line 493, in send_with_data_passthrough
    self._process_event(self.our_role, event)
  File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\h11\_connection.py", line 242, in _process_event
    self._cstate.process_event(role, type(event), server_switch_event)
  File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\h11\_state.py", line 238, in process_event
    self._fire_event_triggered_transitions(role, event_type)
  File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\h11\_state.py", line 251, in _fire_event_triggered_transitions
    raise LocalProtocolError(
h11._util.LocalProtocolError: can't handle event type ConnectionClosed when role=SERVER and state=SEND_RESPONSE
```

w-e-w:
to be honest I don't consider myself very adept at gradio my self
currently I'm trying to make it work as supposed to try to make it look good
the print thing is more of a hotfix
I don't know the proper way of doing this my self 
is the proper way of doing it the method I tried before failed
I have a feeling that the browser has to initiate the request for the element to be updated possibly using some JavaScript

contentis:
The issue you encountered with yield is nothing I have ever seen in our internal testing... Would you mind sharing your OS, Gradio and SD WebIUI version?

---
> I'm not sure if dynamically updating is possible but what I am doing now is rewriting the entire block as a single markdown`
> and we should be able to just update the contents of the markdown when the refresh button is pressed

But this will eliminate the option to have the accordions.

Please note that when it comes to changes to the UI/UX, I won't be making the call.

BurnZeZ:
What exactly are the preinstallation requirements?
Perhaps I ran into a bug but if all necessary dependencies were to be installed automatically something was missing as I had complaints of missing DLLs.
I messed with the Windows’ path to see if I could determine what I was missing (I am tired) but finally that ended with this:
        class TQDMProgressMonitor(trt.IProgressMonitor):
    AttributeError: module 'tensorrt' has no attribute 'IProgressMonitor'

contentis:
@BurnZeZ did you have tensorrt previously installed? Or have tensorrt on your path?

w-e-w:
> But this will eliminate the option to have the accordions.

@contentis no actually example https://github.com/NVIDIA/Stable-Diffusion-WebUI-TensorRT/pull/4

contentis:
Ideally this should be discussed in the repository as issue as this is not part of the PR. And people with the similar issues might get something out of it as well.

TensorRT 9 should be installed by the extension automatically. Can you check pip freeze? There shouldn't be the need to download tensorrt not having it on the path.

w-e-w:
@contentis just to make you aware, I only add extension to the index if they are in working state
(whether or not it gets broken in the future is another matter)
so I don't intend to add this TRT to until at least the "breaking" issues that I found are fixed
> unless the issues I found are due to errors on my part

contentis:
I fully understand that. In the above case I wouldn't consider it a bug / not working. As this is a version mismatch that needs to be handled by the user. And suggesting to move this to an issue is fair in my opinion.

Generally it is arguable what is considered working, due to its hackable nature it is hard to provide a working guarantee.
Setting this up on a clean install, with the recommended python version in a clean environment, has been working on multiple machines.

The main issue I see at the moment is the naming collision with the existing plugin. 

This being said, I'm really happy for your inputs and would love to make this an awesome extension.

BurnZeZ:
> Ideally this should be discussed in the repository as issue as this is not part of the PR. And people with the similar issues might get something out of it as well.
> 
> TensorRT 9 should be installed by the extension automatically. Can you check pip freeze? There shouldn't be the need to download tensorrt not having it on the path.

Pip freeze demonstrated that it believed tensorrt was installed within the original automatic1111 tensorrt extension directory (stable-diffusion-webui\extensions\stable-diffusion-webui-tensorrt\TensorRT-8.6.1.6) and after “uninstalling” it (I removed that directory a while ago but maybe the extension had registered it with pip) pip pulled in the correct version.

If other people find similar problems after having previously used the other extension: The solution is uninstalling tensorrt from pip to clear the stale/outdated reference.

w-e-w:
@contentis 
from my perspective I will consider the `hot_fix` branch to be "working"
ping me when the changes is pushed to master and it's okay for me to add this to index

contentis:
Will do, trying to get some more fixes in that have been reported in the last 24h. Especially addressing the installation process and documentation.

w-e-w:
did a test and didn't explode so hurray

contentis:
> did a test and didn't explode so hurray

I see this as an absolute win
---

Thank you for your support and input!

PR #206, opened 2023-10-13
add sd-webui-FaceChain

wenmengzhou:
## Info 
add FaceChain: https://github.com/modelscope/facechain.git

## Checklist:

- [y] I have read the [`Readme.md`](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions)
- [y] The description is written in English.
- [y] The `index.json` and `extension_template.json` have not been modified.
- [y] The `entry` is placed in the `extensions` directory with the `.json` file extension.

### DIFF
--- a/extensions/sd-webui-FaceChain.json
+++ b/extensions/sd-webui-FaceChain.json
@@ -1,0 +1,11 @@
+{
+    "name": "FaceChain",
+    "url": "https://github.com/modelscope/facechain.git",
+    "description": "FaceChain is a deep-learning toolchain for generating your Digital-Twin. With a minimum of 1 portrait-photo, you can create a Digital-Twin of your own and start generating personal portraits in different settings ",
+    "tags": [
+        "script",
+        "tab",
+        "training"
+    ],
+    "added": "2023-10-13"
+}
--- a/extensions/sd-webui-FaceChain.json
+++ b/extensions/sd-webui-FaceChain.json
@@ -5,7 +5,8 @@
     "tags": [
         "script",
         "tab",
-        "training"
+        "training",
+        "online"
     ],
     "added": "2023-10-13"
 }
--- a/extensions/sd-webui-FaceChain.json
+++ b/extensions/sd-webui-FaceChain.json
@@ -7,6 +7,5 @@
         "tab",
         "training",
         "online"
-    ],
-    "added": "2023-10-13"
+    ]
 }


w-e-w:
did the quick review and it doesn't seem to be working
I don't have time to dig in deeper right now but it seems like you're using diffusers but you also did not didn't install diffusers in install.py
```shell
*** Error executing callback ui_tabs_callback for B:\GitHub\stable-diffusion-webui\extensions\facechain\scripts\facechain_sdwebui.py
      File "B:\GitHub\stable-diffusion-webui\modules\script_callbacks.py", line 166, in ui_tabs_callback
        res += c.callback() or []
      File "B:\GitHub\stable-diffusion-webui\extensions\facechain\scripts\facechain_sdwebui.py", line 15, in on_ui_tabs
        import app
      File "B:\GitHub\stable-diffusion-webui\extensions\facechain\app.py", line 18, in <module>
        from facechain.inference import preprocess_pose, GenPortrait
      File "B:\GitHub\stable-diffusion-webui\extensions\facechain\facechain\inference.py", line 10, in <module>
        from diffusers import StableDiffusionPipeline, StableDiffusionControlNetPipeline, ControlNetModel, \
    ModuleNotFoundError: No module named 'diffusers'
```

w-e-w:
@wenmengzhou hello any updats?

diffusers needs to be added to install.py

---
about onnxruntime
https://github.com/modelscope/facechain/blob/main/install.py#L19

the two versions of onnxruntime, `onnxruntime` `onnxruntime-gpu`
some extensions use one some use the other in general I believe most people would prefer to have `onnxruntime-gpu` version
as you can run on CPU with GPU version but not the other way around
https://github.com/aria1th/sd-webui-nsfw-filter/pull/2 as far as aware this is the best method of installing onnxruntime

---

and also just like you guys have wrote https://github.com/modelscope/facechain/blame/main/install.py#L25 # Todo 这里有坑
I'm also facing trouble installing mmcv
if I'm correct mmcv doesn't seem to provide binaries for python 3.10 so it requires it to be built form sorce
the issue is most people won't have the proper environment to build thus it fails to install
the postal solution is solved is is to provide a binary or write instructions on how to prepare the build environment

wenmengzhou:
@w-e-w  diffusers is added to install.py and onnx-runtime gpu have been supported.  currently mmcv 2.0 is not supported so user have to use older Python version to use prebuilt package or build it from source

w-e-w:
[from webui current dev brach 464fbcd92118bf00173b9982325fe6348201313e](https://github.com/AUTOMATIC1111/stable-diffusion-webui/tree/464fbcd92118bf00173b9982325fe6348201313e) 
pip [freeze clean install windows 3090 without-xformers.txt](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/files/13229403/freeze.clean.install.windows.3090.without-xformers.txt)

wenmengzhou:
https://github.com/modelscope/facechain/pull/403/files  I have checked requirements for FaceChain and cherry pick the necessary requirements for pure sd-webui env

w-e-w:
@wenmengzhou since my fix to the installation process has been merged
I believe this extension is suitable to be added to the index
but there is one last issue that we need to discuss

one of the [rules](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions#tags) of listing an extension on this index is the
> `ads` tag is Required for any extension that contains advertisements or self-advertisement in the extension itself.

this `Github star it here` falls under this rule
image

we do so because to keep the UI clean do not wish people to use this as an explicit advertising platform
if this rule isn't enforced people might start putting random links in strange places
otherwise people would start putting random advertising links all over the place
we believe that the extensions tab is sufficient to direct the user to the extensions repository
as such they should be no reason for a extensions add additional links to do so
image



### the effects of having the `ads` is huge
webui by default will not show an extension listing on the extension tab if a extension have the `ads` tag
the user will have to unchecks `Hide extensions with tags` `ads` in order for them to see it (and most users will likely not do that)
image

here is a example of modification on what is the minimum I believe is acceptable to not required ads tag
image
I've implemented I've implemented this in a way that this should only affect the interface when it's running as a webUI extension
https://github.com/modelscope/facechain/compare/main...w-e-w:hide-self-ad-when-running-as-webui-extension?expand=1

I have to make it clear that removing the self-advertisement is not in requirement of the extension being listed
it is just the requirement of it being listed without the ads tag
you can choose to keep the self advertisement but ads tag the result will be hidden by default
or you can remove the self-advertisement part in listed as a normal extension

if you wish to remove it you can pull use my implementation or do it some other way

note1:
there's also other links in the extension
`Paper cite it here` because the link is huge it might also be classified by an advertisement
currently my stance is giving this a "pass" because the use case is a bit different, but in the future if people makes incomplete about this then I might need to come back and add the ads tag
ideally the link is also removed or at least moving to the bottom of the page

note2: The Links at the bottom of the page which also serves the same purpose as linking to the original repo
I considered this has been documentation used
I also give those things a pass but again, if people makes incomplete about this then I might need to come back and add the ads tag

please tell me you decision on this matter

wenmengzhou:
sure,  for Note1 & Note2  I will discus with other developers and reply it asap.   As for Note3, I do not find a link in the claim, which link do you mean?

w-e-w:
> As for Note3, I do not find a link in the claim, which link do you mean?

that's a misunderstanding, I'm not saying the the disclaimer is it advertisement link
I'm just saying that I have no objections with the disclaimer

w-e-w:
more details
> sorry if I repeat myself

github repo link present it in the way that cannot be interpreted as a documentation link is considered as an advertisement

this is an exception for documentation links (such as the one you have at the bottom of page) as long as the extension is sufficiently complex, which list extension is.

we believe that the extension tab already provide clear and uniform place to find links to repo
that's why github repo link are classified self-advertisement

the only links are of concern is the ones in the application itself, the is not restrictions with readme or documentation

---

I believe that citation links can be exempted from `ads` tag, since the use case is different
but due to the links size and position of your citation link, this is a bit on shaky grounds
so if someone were to make a complaint the judgment on this matter to may have to change
> in my opinion if the link is normal size and at the bottom then I think no one will have any issues with it

---

please understand I'm not trying to make things difficult
I'm trying my best to maintain a clean environment for webui

since I'm currently the only one thats actively maintaining the index I have to be extra careful about how the rules are interpreted
if you believe that my judgment on this matter is poor, I could try and get other people to also involve in this discussion
aslo if compelling reasoning is presented rules can be changed

w-e-w:
@wenmengzhou 
I just noticed that you pulled my example branch https://github.com/modelscope/facechain/commit/47a7e5052fa01e2466844f12be2a1781236c147b
I take this as confirmation that this extension can be added without the `ads` tag

### there's just one more issue (hopefully the last issue) that I need to make sure before this can be added to index

I was doing some testing and it seems to me that this extension require the `online` tag
the online tag is required for an extension that requires connection to an external server to function (more explanation below)

---

### **do not worry**
this is unlike the ads `tag`, the `online` tag is just a "normal" tag, they won't be any issues of an extension being "hidden", it is used to inform a user that extension relies on external server, and to inform those that have privacy concern they may not want to use this extension

---

### and extension is required to have the `online` tag if
- requires an external server function
- uploads user data to an external server
- performs non one-time downloading asset

exceptions to the online tag
- extensions that performs one-time downloading of assets
-  - if an extension requires an asset to function such as a model, it will make a one-time request to download model, the model will be cached and reused, it should be able to run offline without a internet connection if the asset has already been downloaded
- - - example of this can be seen in how webui downloads image upscaling their models, webui makes a request to download the model if the model is not already downloaded
- it only exclusively connects to a very well know server like GitHub to perform non privacy infringing actions like update check

---

during my testing during my testing of "training a face", FaceChain almost falls into the exception of `one-time downloading of assets`, as it downloads and cache the model for later use, but for some reason it seems to make a request to a server to confirm the location of the model before it actually use the cached model, and when connection is not available training is not ableable to function properly

as this extension fails to function when internet connection is not present
and seems to use and seems to utilize cookie doing the API request with server which can possibly be considered as uploading trackable data, this extension requires the `online` tag

can you come from analysis is correct
if so plese  add the `online` tag to the extension entry
> also plese remove the "added": "2023-10-13", the date is automatically generated upon merge otherwise the date is incorrect

if future the situation has changed you make a pull request to have the `online` tag removed
you also welcome to update the description in the future if necessary

---

apart from the above `online` tag issue, I don't believe there's any issues
I will be able to add the extension to index after the above situation has become confirmed

hopefully in the future mmcv installation will be more streamlined for the average user

wenmengzhou:
thanks,  we have discussed the github  link issue and merged your pull request in facechain.

your analysis about online is correct, models will be downloaded automatically for the first time,  and after that each time we use the model, modelscope sdk will check whether the local model is in consisitant with the server, so network is required now, I will add online tag today

w-e-w:
thank you for understanding patience and cooperation, I will merge this into the index now

I have removed the incorrect added date, the date is automatically generated upon merge

2blackbar:
This is not working, cause of mmcv , pip install mmcv-full wont solve the mmcv error, do we have to become python devs to use this one ? Im on 3090 

`Traceback (most recent call last):
  File "C:\Users\BAZO\AppData\Local\Programs\Python\Python310\lib\multiprocessing\process.py", line 314, in _bootstrap
    self.run()
  File "C:\Users\BAZO\AppData\Local\Programs\Python\Python310\lib\multiprocessing\process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "H:\sd\extensions\facechain-fix-sdwebui_req\facechain\inference.py", line 24, in _data_process_fn_process
    Blipv2()(input_img_dir)
  File "H:\sd\extensions\facechain-fix-sdwebui_req\facechain\data_process\preprocessing.py", line 207, in __init__
    self.face_detection = pipeline(task=Tasks.face_detection, model='damo/cv_ddsar_face-detection_iclr23-damofd', model_revision='v1.1')
  File "H:\sd\venv\lib\site-packages\modelscope\pipelines\builder.py", line 164, in pipeline
    return build_pipeline(cfg, task_name=task)
  File "H:\sd\venv\lib\site-packages\modelscope\pipelines\builder.py", line 67, in build_pipeline
    return build_from_cfg(
  File "H:\sd\venv\lib\site-packages\modelscope\utils\registry.py", line 215, in build_from_cfg
    raise type(e)(f'{obj_cls.__name__}: {e}')
ModuleNotFoundError: FaceDetectionPipeline: DamoFdDetect: No module named 'mmcv'
instance_data_dir H:\sd\extensions\facechain-fix-sdwebui_req\worker_data\qw\training_data\ly261666/cv_portrait_model\person1`

w-e-w:
@2blackbar please refrain from using this repository for help

PR #219, opened 2023-10-25
Add sd-webui-facefusion

linjiX:
## Info 
https://github.com/diffus-me/sd-webui-facefusion

## Checklist:

### DIFF
--- a/extensions/sd-webui-facefusion.json
+++ b/extensions/sd-webui-facefusion.json
@@ -1,0 +1,9 @@
+{
+    "name": "sd-webui-facefusion",
+    "url": "https://github.com/diffus-me/sd-webui-facefusion",
+    "description": "Video face swapper and enhancer.",
+    "tags": [
+        "tab",
+        "editing"
+    ]
+}


w-e-w:
I did a quick test and the extension did seem to work for me

maybe I should rewrite the description so it's more easily understandable
https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions#tags

> `ads` tag is Required for any extension that contains advertisements or self-advertisement in the extension itself.

image

even though the links are not your link this still supposed under require the `ads` tag
having the `ads` tag ment it will not be showned automatically on the index page
one will have to uncheck the filter
image
it is up to the developer to either have links in the extension and have the `ads` tag or remove the links and not have the ads tag
face with this decision most developers choose to remove ads from the extension
is your choice but in this case I don't know it will be right for you to do so as I believe you basicly ported the original repo
I would suggest you consult the original developer on this matter

w-e-w:
other things that would cause issues
in your install.py you strictly specified which versions of packages you wish to install
some of the packages you specified to be installed are overriding web UI default, since when do I would install its packages at lunch this means on every launch time will be wasted on reinstalling packages

moreover you have not considered any extension compatibility, as different extensions we have different requirements your extension will have the potential of breaking other extensions

I believe the best practice in installing packages is only is the only specify a minimum package version
and if some packages is already installed by web UI don't install it again especially don't override it

on the only topic of packages, about onnxruntime, as they exist onnxruntime and onnxruntime-gpu
I belive it is best to test which version you can install and install that version
I'm not sure what will happen if you try and install the GPU version on a non GPU system
example https://github.com/w-e-w/sd-webui-nudenet-nsfw-censor/blob/main/install.py

I've done a quick test with facefusion and it performs much better with onnxruntime-gpu running on CUDA compared to CPU

w-e-w:
dont import torch there
https://github.com/diffus-me/sd-webui-facefusion/blob/main/install.py#L7
see https://github.com/aria1th/sd-webui-nsfw-filter/pull/2

w-e-w:
yes I see you remove the links
just to make it clear it is okay to put donation links in the readme there's no restriction of ads tag on that only in the extension itself

---

the following issues can be solved later so I'll add this to index now

note as I am typeing this I saw [build: lower the required version of "numpy" and "protobuf"](https://github.com/facefusion/facefusion/commit/315e3bca68827e6dd385771bc2f48847034375af)
so some of these stuff maybe irrelevant now

---

the basic idea is if web UI install something by default you should not need to install it yourself unless absolutely necessary

stuff like
numpy>=1.24.3
pillow>=10.0.1
at best this would cause the package to be constantly installed and reinstalled on every launch of web UI
```
Installing sd-webui-facefusion requirement: changing numpy version from 1.23.5 to 1.24.3
Installing sd-webui-facefusion requirement: changing pillow version from 9.5.0 to 10.0.1
Installing sd-webui-facefusion requirement: changing protobuf version from 3.20.3 to 4.24.2
```
at worse this might cause things to break
> luckily this doesn't seem to be the case

> for Pillow hopefully we can get Auto too upgrade to 10.0.1 but as of now webui is on 9.5.0
there's some bug fixes I want in 10.0.1
and blendmodes dependency is upgreade to 10.0.1 in ver 2023, but the current 2022 >= 9.0.0, <10

for more webui package version see https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/dev/requirements_versions.txt
note: they are lots of packages not listed in requirements_versions.txt but are still installed because they are dependencies of the install packages such as for example opencv

pipdeptree output
```
* blendmodes==2022
 - Pillow [required: >=9.0.0,<10, installed: 10.0.1]
* open-clip-torch==2.20.0
 - protobuf [required: <4, installed: 4.24.2]
```
hopefully these are just pip complaining and not real dependency issues

using `pipdeptree` it doesn't seem anything requires `numpy>=1.24.3` so you should probably not specify it

---

suggestion for future
due to this being a port of the original code
the UI is not written in the conventional web UI method
this causes certain things that should automatically work in web UI to not work
for example webui has a `ui-config.json` / `Settings > Defaults` system, this system allows the use to change a UI default options to their liking, example this allows the user to set the default imagination size to be something other then the default 512x512
if the UI is written in the conventional webui way it will allow one to set the default provider to CUDA

---

also potential upcoming issues not sure what the cause and this could be only just on my system
I have encountered an issue that when calling `predict_video()` it triggers an error
```
opencv-python\opencv-python\opencv\modules\highgui\src\window.cpp:1266: error: (-2:Unspecified error) The function is not implemented. Rebuild the library with Windows, GTK+ 2.x or Cocoa support. If y
ou are on Ubuntu or Debian, install libgtk2.0-dev and pkg-config, then re-run cmake or configure script in function 'cvDestroyAllWindows'
```
I solve this by manually uninstalling and reinstalling `opencv-python`
it's possible that my installation is about corrupted I already have no idea what's going on 
if other people's report this issue you make try asking them to try reinstalling opencv

PR #222, opened 2023-11-01
Add CADS extension

v0xie:
## Info 
https://github.com/v0xie/sd-webui-cads
An implementation of the method described here: https://arxiv.org/abs/2310.17347

## Checklist:

### DIFF
--- a/extensions/CADS.json
+++ b/extensions/CADS.json
@@ -1,0 +1,9 @@
+{
+    "name": "CADS",
+    "url": "https://github.com/v0xie/sd-webui-cads.git",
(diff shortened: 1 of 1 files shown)


w-e-w:
I don't pretend to know what you're doing as I don't have time to read through the paper
all I can say is that it didn't explode in my webui installation so it got a pass

when doing my initial test I have `high res fix` enabled, and let's just say the output is not good
I think you might need some extra logic to deal with `high res fix`

v0xie:
@w-e-w Thanks for the review! I'll look into supporting hi-res fix ~

PR #225, opened 2023-11-02
Add sd-webui-temporal extension

Iniquitatis:
## Info 
[https://github.com/Iniquitatis/sd-webui-temporal](https://github.com/Iniquitatis/sd-webui-temporal)

## Checklist:

### DIFF
--- a/extensions/sd-webui-temporal.json
+++ b/extensions/sd-webui-temporal.json
@@ -1,0 +1,9 @@
+{
+    "name": "Temporal",
+    "url": "https://github.com/Iniquitatis/sd-webui-temporal",
+    "description": "A \"loopback on steroids\" that allows either processing images in a slightly more advanced way, or making the morphing animations out of them.",
+    "tags": [
+        "script",
+        "animation"
+    ]
+}


w-e-w:
you failed to mention that you were extension requires controlnet
please add this into your documentations
> and maybe also the description on the index if you think it's suitable

the method you import controlnet is also wrong
an extension is not necessary installed under the same name
image
and when `--data-dir` is used it won't even be in the `webui_root/extensions/name`

your method of importing control Network only work when the controlnet is installed in the "default directory" under the "default name"

this is how another extension imports contronlet
https://github.com/continue-revolution/sd-webui-animatediff/blob/master/scripts/animatediff_cn.py#L31-L35
only import contronnet after that it's actually needed (in this case during before_process) https://github.com/continue-revolution/sd-webui-animatediff/blob/master/scripts/animatediff.py#L57
this ensures that your extension will always work regardless of extension load order (which currently depends on the extension directory name)

Iniquitatis:
@w-e-w Thanks a lot for the review!

> you failed to mention that you were extension requires controlnet

It doesn't actually require ControlNet per se, but simply supports saving/loading its parameters. Forgot to make the dependency optional, so thanks again for the catch.

> the method you import controlnet is also wrong

I've used an [official example](https://github.com/Mikubill/sd-webui-controlnet/wiki/API#create-controlnet-arguments). Nonetheless, fixed.

w-e-w:
> It doesn't actually require ControlNet per se...

ahh that makes sense I was kind of confused when I was looking for your code

> I've used an official example...

I guess someone will need to update the examples on their Wiki

PR #226, opened 2023-11-03
Add sd-webui-extended-style-saver

harukei-tech:
## Info 
https://github.com/harukei-tech/sd-webui-extended-style-saver

## Checklist:

### DIFF
--- a/extensions/sd-webui-extended-style-saver.json
+++ b/extensions/sd-webui-extended-style-saver.json
@@ -1,0 +1,10 @@
+{
+    "name": "sd-webui-extended-style-saver",
+    "url": "https://github.com/harukei-tech/sd-webui-extended-style-saver.git",
+    "description": "Save your AI prompts easily. This extension helps you keep important details like name, model, VAE, image size, prompt, and negative prompt for creating AI outputs.",
+    "tags": [
+        "script",
+        "UI related",
+        "prompting"
+    ]
+}


w-e-w:
@harukei-tech from my testing your extension fails to function
the error caused by your extension even causes web UI to fail
image
please retest your extension
when testing remember to use a clean install of `webui 1.6.1` and or `dev branch`

<details><summary>Error</summary>
<p>

note this is only the first exception output
```shell
*** Error calling: B:\GitHub\stable-diffusion-webui\extensions\sd-webui-extended-style-saver\scripts\extended_styles.py/ui
      File "B:\GitHub\stable-diffusion-webui\modules\scripts.py", line 402, in wrap_call
        return func(*args, **kwargs)
      File "B:\GitHub\stable-diffusion-webui\extensions\sd-webui-extended-style-saver\scripts\extended_styles.py", line 75, in ui
        self.ui_txt2img()
      File "B:\GitHub\stable-diffusion-webui\extensions\sd-webui-extended-style-saver\scripts\extended_styles.py", line 105, in ui_txt2img
        apply_button.click(
      File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\gradio\events.py", line 140, in __call__
        dep, dep_index = self.trigger.set_event_trigger(
      File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 287, in set_event_trigger
        "outputs": [block._id for block in outputs],
      File "B:\GitHub\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 287, in <listcomp>
        "outputs": [block._id for block in outputs],
    AttributeError: 'NoneType' object has no attribute '_id'

``` 

</p>
</details>

harukei-tech:
@w-e-w 
Thank you for your polite bug report!

Upon checking with a clean-installed version of `dev branch`, I confirmed the 'NoneType' error as you pointed out.

While this issue has been resolved, I discovered another error related to ControlNet in subsequent processes. Resolving this new issue would take some time, I have removed the processing related to ControlNet and will address it another branch.

Since the description in the file extensions/sd-webui-extended-style-saver.json of this PR does not mention ControlNet, nothing was  updated.

Thank you again for your attention and verification.
image

w-e-w:
I will check the extension again when I have time

if the trouble you're having with controlnet is due to load order this might help a bit
https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/225#issuecomment-1792148215

also I'm currently working on the system that wold allow an extension to have total control of the load order every script and the execution order of every callback
note: this may or may not merged into to webUI I haven't passed this through AUTO yet

w-e-w:
nice retested and it didn't explode
>even though in my opinion I would probably use the default built-in styles
>if you just saving type engine info and then click apply it can basically be used similar to this extension but this is just my opinion

tip:
you don't have to actually modify the settings to apply `sd_model_checkpoint` and `sd_vae`
https://github.com/harukei-tech/sd-webui-extended-style-saver/blob/master/scripts/service/extended_style_service.py#L55-L62
instead you can just add it to `override settings`
image
the advantage of doing it this way is that for one it doesn't modify the settings which means that it's better in multi-user use case
using `override settings` basically changed the setting temporarily when the image is generated, so it's guaranteed that it will be color correct model
where are settings is global so in a multi-use scenario if one user change the model then the models will be changed for every one
`override settings` is very powerful almost all settings can be changed with it

harukei-tech:
Thank you for your review and merge.
I also appreciate your good advice! I plan to address each of them.

## ControlNet Support
### Load Order
Thank you for providing the example of the [hack method](https://github.com/continue-revolution/sd-webui-animatediff/blob/master/scripts/animatediff.py#L57). 
That method's process uses external source code so it's hard for me to comprehend right away.
I'll read it again when I have finished responding to the following Issue.

### Current issue in sd-webui-extended-style-saver
This is not related to the load order.

When [gr.Image.update](https://github.com/harukei-tech/sd-webui-extended-style-saver/blob/feature/control_net/scripts/usecase/apply_usecase.py#L34-L36) is performed on the image component of ControlNet,
the gradio framework's [preprocess](https://github.com/gradio-app/gradio/blob/v3.41.0/gradio/components/image.py#L265) executes about 4-5 times with different arguments.

It eventually completes successfully, but a type mismatch error appears in the terminal several times.
It is difficult and time consuming to figure out whether SD, Gradio, or ControlNet should be addressed.

## Override Settings
I wasn't aware of such an option. 😮 
Thanks for the useful information!  

I would like to try incorporating it, but could you please tell me how to display this item? 
Even when I try navigating to "Settings > User interface > txt2img/img2img UI item order", it doesn't appear.
I searched the internet for information on how to display it but couldn't find any.

w-e-w:
override setting doesn't appear in the UI it's not meant to be used by the user directly
it automatically appears when you read PNG info, an override setting entry will appear if it is used,
or use as internal option in extensions or scripts, see the built-in XYZ script, in there I use override settings to apply lots of options

PR #229, opened 2023-11-10
Added sd-webui-modal-info.json extension information.

luminouspear:
## Info 

Updated modal overlay to show prompt information and some other consistency/cleanup updates in the CSS.

## Checklist:


w-e-w:
while testing I found that your extension makes the image browsing experience very inconsistent

issues like 
- only be able to enter in gallery mode once, after exiting it will not be able to enter it again 
- not able to see the prompt overlays you describe in the documentation

maybe I'm doing something wrong, I have tested in both master and dev version of web UI
sorry to tell you that my assessment is that this extension cannot be listed in the index at its current state

---
from your readme [Known Issue](https://github.com/luminouspear/sd-webui-modal-info#known-issues)
> When the total number of images generated (calculated as batch size multiplied by the number of batches) is less than during the previous generation, the application's client-side behavior leads to the deletion of the image preview div. This, in turn, removes the previous thumbnails and disables the modal.


calculating the total Image number based on batch size and batch number is probably not a good idea
things like scripts and extensions can change the number of image output, examples XYZ grid or img2img batch processing, they're also numerous extensions that retrns extra images in the gallery

luminouspear:
https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/assets/61656433/bea80c20-08d9-4325-91d8-de564af42a5a

Wow, yeah I wouldn't add it if I was having those issues either. This is how it works for me. Tested in both Chrome and Firefox, unsure why you're having those issues. Definitely open to better ways to get the count of elements (I'm counting the number of thumbnails available in the thumbnail gallery and watching it for changes).

w-e-w:
well something is definitely broken in your js

https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/assets/40751091/a5d380f1-b4ed-4137-94cb-d3257e6515f7

luminouspear:
I think the issue was the empty negative prompt. I have added code to account for all the different prompt states, should be working now.

w-e-w:
did a quick test and yes it's working
I will merge this now

luminouspear:
> did a quick test and yes it's working I will merge this now
Awesome, thanks for the video, too, very helpful.

> just a IMO, it's totally your decision but I personally wouldn't put None when there's nothing to display as an edge case would be someone would have the literally `None` as the negative prompt
As it sits now if someone has `None` as the negative prompt then `None` will be output, if they have `` as the negative prompt then `None` will be the output. I tried leaving it empty but it looked more off, but I will take it as a suggestion for sure, thank you.

PR #232, opened 2023-11-18
add style-vars extension

SirVeggie:
## Info 
https://github.com/SirVeggie/extension-style-vars

## Checklist:


w-e-w:
nice, works as intended without issues

---

just a suggestion
I think it makes more sense to put your enable option in setting instead of a accordion 
a user is not expected to change this option all the time
this way it declutter's the UI

SirVeggie:
That's a good point, I moved the toggle to the settings tab. 👍

PR #240, opened 2023-12-01
Add sd-webui-breadcrumbs extension

ThereforeGames:
## Info 
Repo: https://github.com/ThereforeGames/sd-webui-breadcrumbs

## Checklist:

### DIFF
--- a/extensions/sd-webui-breadcrumbs.json
+++ b/extensions/sd-webui-breadcrumbs.json
@@ -1,0 +1,8 @@
+{
+    "name": "sd-webui-breadcrumbs",
+    "url": "https://github.com/ThereforeGames/sd-webui-breadcrumbs.git",
+    "description": "Adds a breadcrumb trail and makes improvements to the Quicksettings menu.",
+    "tags": [
+        "UI related"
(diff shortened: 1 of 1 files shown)


w-e-w:
image

nice, I think this can be considered as a documentation link as it is not in your face

---

> Use custom CSS for scrollbars May require a full restart or a hard refresh to take effect.)

are you sure that a full restart is required? as I recall most CSS and even JS stuff should update immediately after browser reload (F5), at least a reload with no cache (Ctrl + F5)

---

~~by the way is breadcrumbs a industry term for this type of UI?~~

~~hmm, who would have guessed, I can answered my question by searching 5 seconds, yes it is apparently it is~~

w-e-w:
also just so you're aware your other extension `Unprompted` also have the `ads` tag because of well those sponsor / donation links and stuff

and again it is your free choice to include them or not

if you decide to remove those elements we can update the listaing and have the tag removed

note: I wasn't the one who added `Unprompted` to index in the back then, so I'm not sure if you're aware that it has the tag and the implications of it

PR #244, opened 2023-12-07
add Color Correction Extras

light-and-ray:
## Info 
https://github.com/light-and-ray/sd-webui-color-correction-extras
This simple extention adds native color correction feature from stable-diffusion-webui into "Extras" tab
So, you can use this feature in postprocessing

## Checklist:

### DIFF
--- a/extensions/Color Correction Extras.json
+++ b/extensions/Color Correction Extras.json
@@ -1,0 +1,9 @@
+{
+    "name": "Color Correction Extras",
+    "url": "https://github.com/light-and-ray/sd-webui-color-correction-extras.git",
+    "description": "Adds native color correction feature into \"Extras\" tab",
+    "tags": [
+        "manipulations",
+        "extras"
+    ]
+}
--- a/extensions/Color Correction Extras.json
+++ b/extensions/Color Correction Extras.json
@@ -3,7 +3,6 @@
     "url": "https://github.com/light-and-ray/sd-webui-color-correction-extras.git",
     "description": "Adds native color correction feature into \"Extras\" tab",
     "tags": [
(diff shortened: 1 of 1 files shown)


w-e-w:
facepalming, not sure why we didn't think of including color correction into the post processing tab

I remove the manipulation tag because this does not fit
https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/blob/32506e916ea5c36fd1d319e2729b07b7a8f41d09/tags.json#L13

w-e-w:
> @w-e-w if you add similar feature in future update, can you add more then 1 method, like additional 2 from StableSR extensions? It's looking like AdaIN produces similar and better results to a1111, but much less dithering; and Wavelet mixes pictures, so it can be better it case pictures are similar

I don't understand what you're saying

w-e-w:
if  by you "you" he actually mean "you" as in "I"

if you update the extension in the future to add more color correction methods is your choice nothing to do with us
as long as it's still working extension we don't really care much

if you wish to update the description of this extension, just make a new PR
if you put ads in the extension that your required to add ads tag
if you start using online services then the online tag is required

w-e-w:
ahh see what you mean

I guess the original reason why we didn't implement it into post-processing extra tab is because as you have noticed it requires a second input, back then we probably don't want to complicate the UI (it's already very complicated)

while we could Implement new and better color correction methods
but if I were to add it I will probably also add it as an extension
not only that this is easier to add, improved and updated and not limited to webui's update schedule
furthermore this type of things might require extra dependencies and this would make webui harder to maintain

I think it makes better sense like what you did here for those who needed at it as extension

PR #201, opened 2023-10-04
add sd-webui-workspace

Scholar01:
## Info 

Designed to save and restore Webui workspace. It can quickly save your current webui interface and restore it the next time you use it.

## Checklist:

- [x] I have read the [`Readme.md`](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions)
- [x] The description is written in English.
- [x] The `index.json` and `extension_template.json` have not been modified.
- [x] The `entry` is placed in the `extensions` directory with the `.json` file extension.

PR #248, opened 2023-12-14
Add state-manager to the index

SenshiSentou:
## Info 
https://github.com/SenshiSentou/sd-webui-state-manager

I did add the 'tab' tag, even though technically it's its own panel, unrelated to the gradio tabs. Not sure if that is correct or not.

## Checklist:


w-e-w:
I'm a bit torn on this one
as you point out in your readme your extension is very WIP and unstable
seems to be "working" in the current master RC and dev, but it does spits out tons of warnings and errors in the browser console

since is technically working I think this extension can be added to the index, but I don't feel too comfortable about it, hopefully you can make it more stable in the future

---

not sure if this would help or if you're aware, in webui there is a ui-config.json and the defaults system, it essentially allows the user to configure the default values of elements, if you don't know about the system in web UI then this may be of use to you, 
> see ui-config.json and `settings > defaults > view caanges/apply`

> yes I realize that you were extension works totally in the front end, while the defaults system work form the backend

and in case you are not aware, webui currently in the progress of upgrading to gradio 4.x https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/14184 we haven't decided on when this will be added to web UI but definitely not in 1.7 
(likely in a 1.8 or 1.9) but you might want to keep track on this

SenshiSentou:
> I'm a bit torn on this one as you point out in your readme your extension is very WIP and unstable seems to be "working" in the current master RC and dev, but it does spits out tons of warnings and errors in the browser console
> 
> since is technically working I think this extension can be added to the index, but I don't feel too comfortable about it, hopefully you can make it more stable in the future

Thanks, I appreciate you adding it. I would absolutely love to find a more robust way of handling this, but the lack of consistent ids/ named attributes makes this ridiculously challenging. 

> and in case you are not aware, webui currently in the progress of upgrading to gradio 4.x [AUTOMATIC1111/stable-diffusion-webui#14184](https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/14184) we haven't decided on when this will be added to web UI but definitely not in 1.7 (likely in a 1.8 or 1.9) but you might want to keep track on this

I was not aware, but that's awesome! I haven't looked into it too much, but I remember reading an issue a while ago that talked about gradio exposing more functionality in JS-land?

Also, is there like a workgroup, or some sort of centralized development talk for A1111? I've looked around before and it'd be awesome to be in the loop more.

> not sure if this would help or if you're aware, in webui there is a ui-config.json and the defaults system, it essentially allows the user to configure the default values of elements, if you don't know about the system in web UI then this may be of use to you,
> 
> > see ui-config.json and `settings > defaults > view caanges/apply`
> 
> > yes I realize that you were extension works totally in the front end, while the defaults system work form the backend

I'll have a look at that, thanks! I've looked at various different avenues for getting and manipulating the data from the frontend, but didn't really find anything too useful for this purpose.

w-e-w:
> I was not aware, but that's awesome! I haven't looked into it too much, but I remember reading an issue a while ago that talked about gradio exposing more functionality in JS-land?

yes gradio 4 add adds official custom JavaScript support, and it's possible to write your own custom elements now
>they also change the event function `_js` to `js`, because well it's now officially supported
which means that we have to insert a conversion layer in webui to make `js` -> `js`

SenshiSentou:
Awesome, thank you!

The custom components will be great, but I'm hoping there can also be an improvement in non-custom/ generated UI elements. For example, the `id`s of parameters in scripts and accordions are just `component-###`. Obviously this is coming from a relatively niche use-case, but in general I think it'd be nice if these generated components had:

- Descriptive IDs or `data-setting` attributes e.g.
- A way of exposing their data and allowing it to be updated in JS (I'm imagining something like a `injectedScripts['cutoff'].set('enabled', true)`, which leads to...)
- A proper way of managing and accessing injected scripts and components, pythonic or otherwise. So many JS extensions just pollute the global namespace for example, it would be great if there was a standardized mechanism for registering and accessing extension objects.

And since I'm rambling about structure and JS-land, maybe we could have a more robust way of injecting custom elements? This couldn't possibly hope to cover all the bases, but even just a few commonly-used ones like a `quickSettings.add(myButton)` would be better than a `gradioApp().querySelector('#name')`.

I really am rambling now, sorry for going so off-topic! But I'm really excited to watch A1111 grow, and I think some structure would go a very long way

w-e-w:
I remember the other state extensions
maybe you can grab some ideas from there
https://github.com/ilian6806/stable-diffusion-webui-state
https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/201

SenshiSentou:
I did check alternatives as well, but didn't get much father. The webui-state you linked above also depends on element order and label text[1]. Workspace I hadn't seen before! A quick glance seems to suggest they're serializing on the python side, which would be interesting for sure!

[1]
```js
let label = checkbox.nextElementSibling;
let id = state.utils.txtToId(label.textContent);
let value = store.get(id);
```

w-e-w:
well good luck

---

pointing out the obvious, web UI is it open source project
which means within reason you can have something added to web UI if it would help you make an extension work
for example they have been multiple cases of extensions requesting the additions of new callbacks

I don't know what you would need to get your extension working but if you do it doesn't necessarily have to be totally contained within the extension

SenshiSentou:
I did briefly consider that, so good to know that that's a viable option. I'll look at that and the python helper later this week, thanks for all the help!

w-e-w:
no problem
also considering joining the auto's dev discord
you may be able to get some help if you need it
invitation think on the wiki https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Contributing

PR #258, opened 2024-01-08
add Latent Regional Helper

safubuki:
## Info 
<https://github.com/safubuki/sd-webui-latent-regional-helper>

## Checklist:

### DIFF
--- a/extensions/sd-webui-latent-regional-helper.json
+++ b/extensions/sd-webui-latent-regional-helper.json
@@ -1,0 +1,9 @@
+{
+    "name": "Latent Regional Helper",
+    "url": "https://github.com/safubuki/sd-webui-latent-regional-helper.git",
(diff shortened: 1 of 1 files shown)


w-e-w:
I don't use [regional-prompter](https://github.com/hako-mikan/sd-webui-regional-prompter) so I don't know if it actually functions as intended but at a glance it does seems to generate the division ratio in the correct format
and seems to work without issues so nice

just one thing nothing critical, best practice is to not have at all `install.py` if you're not using it
reason being that install.py is executed under a different sub process
and if the file exists webui will waste a bit of time executing an empty file doing useless work

image

image

> 0.181 seconds wasted for nothing

safubuki:
@w-e-w 
Thank you for sharing with me the best practices for extensions.
I am very happy with your help.

I immediately fixed my extension with your advice.
And the startup time is now 0sec.
image

w-e-w:
nice

PR #263, opened 2024-01-13
sd-webui-thumbnailizer

MNeMoNiCuZ:
## Checklist:

### DIFF
--- a/extensions/sd-webui-thumbnailizer.json
+++ b/extensions/sd-webui-thumbnailizer.json
@@ -1,0 +1,11 @@
+{
+    "name": "Thumbnailizer",
+    "url": "https://github.com/MNeMoNiCuZ/sd-webui-thumbnailizer.git",
+    "description": "A thumbnail gallery and set management tool. Generate and switch between different thumbnail sets for your checkpoints.",
+    "tags": [
+        "thumbnail",
+        "gallery",
+        "generate",
+        "tab"
+    ]
+}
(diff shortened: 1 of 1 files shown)


MNeMoNiCuZ:
Added sd-webui-thumbnailizer.json

w-e-w:
some issues 
your default `set.json` it's also used as the users set.json
if you update to the default in the future I think this will cause the users to be overwritten
I recommend you separate users and default it into different files

also one thing I think users are going to ask
you're not writing infotext to the image metadata, which means that 
if you set the seed to be random (which you have for the `Preview` profile) you won't be able to generate the image again because the information seed info is lost

MNeMoNiCuZ:
> some issues your default `set.json` it's also used as the users set.json if you update to the default in the future I think this will cause the users to be overwritten I recommend you separate users and default it into different files
> 
> also one thing I think users are going to ask you're not writing infotext to the image metadata, which means that if you set the seed to be random (which you have for the `Preview` profile) you won't be able to generate the image again because the information seed info is lost
> 
> these does not prevent the extension from working

Thanks for the feedback!
I didn't consider the default set.json and user.json.
That's a very good point. I guess the safest way to handle it is to do some initialization the first time the script is run. If a user_sets.json exists, use that. If not, copy default_sets.json as user_sets.json?
This way I can still update defaults for new users.

I couldn't figure out how to add the metadata. I would love to do this as well. Do you have any place where I can read some code to figure it out?

Thanks for the review, much appreciated! @w-e-w

PR #270, opened 2024-01-22
sd-webui-inpaintconverter

groinge:
## Info 
[https://github.com/groinge/sd-webui-inpaintconverter](https://github.com/groinge/sd-webui-inpaintconverter)

Simple extension that removes the need for discrete inpainting models. 

The extension downloads this model when first enabled. Which is a unet from the merge v1.5inpainting - v1.5
https://huggingface.co/groinge/inpaintdifferencemerge/commits/main/V1_5inpaintdifference.safetensors

## Checklist:

### DIFF
--- a/extensions/sd-webui-inpaintconverter.json
+++ b/extensions/sd-webui-inpaintconverter.json
@@ -1,0 +1,8 @@
+{
+    "name": "Inpaint Model Converter",
+    "url": "https://github.com/groinge/sd-webui-inpaintconverter.git",
+    "description": "Automatically converts any SD v1.5 checkpoint in to a temporary inpainting model",
+    "tags": [
+        "models"
+    ]
+}


w-e-w:
cool you're I belive basically doing [this](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/How-to-make-your-own-Inpainting-model) but automatically correct?

I haven't actually run your your code yet so I'm not sure if it actually works but assuming that it does
but from code reading I found a some potential issue

basically models can be switched after `setup()`
the most common reason `override setting`
they also other things such as XYZ grid and Refine (yes you can refine with sd 1.5) not to mention other extensions that switch models for whatever reason

I think it's better that if you hijack module loading / switching otherwise they might be various reasons that models might be switched

also another thing not an issue but I think you would want to add using converted in paint model to infotext

groinge:
I didn't really consider that. Never used plotting or a refiner with inpainting but yeah I should try to support it.
I did work on something similar with a another extension I was working on, but I couldn't find a consistent way to detect a model change. Model_loaded_callback works for most things except for when sd_models.reload_model_weights reloads an already loaded model, which ultimately makes it kinda useless for what I'm doing.

w-e-w:
there's also a couple other things that you might want to watch out

image

image

I have a feeling that leaves two things might cause issues 

---

anther I think should be fine but I think it's better that you confirm again
make sure that 
```py
checkpoint_info.name_for_extra = f"temp-inpainting-{checkpoint_info.name_for_extra}"
```
will make the checkpoint be "recognized" as different checkpoints when calculating conds

as webui caches the conds if it doesn't need to be recalculated
https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/94d4b3c8e78df41eea431069475e1d35f75d02ff/modules/processing.py#L441-L491
if for whatever reason that your modification would result in it being recognized at the same model, then the conds it may not be recalculated when required which will cause issues and producing erroneous results

w-e-w:
- https://github.com/groinge/sd-webui-inpaintconverter/pull/1

groinge:
Hey thanks for the help. But after reflecting a bit I think I will just close this. I'm very inexperienced so I don't think I'd be able to keep this repo to a high enough standard for such a wide release.

w-e-w:
> Hey thanks for the help. But after reflecting a bit I think I will just close this. I'm very inexperienced so I don't think I'd be able to keep this repo to a high enough standard for such a wide release.

if you change your mind let me know

also I might have a go if I have time

PR #275, opened 2024-02-08
sd-webui-creaprompt

tritant:
## Checklist:

### DIFF
--- a/extensions/sd-webui-creaprompt.json
+++ b/extensions/sd-webui-creaprompt.json
@@ -1,0 +1,9 @@
+	{
+    "name": "sd-webui-creaprompt",
+    "url": "https://github.com/tritant/sd-webui-creaprompt.git",
+    "description": "This extension for A1111 generates wonderful prompts randomly",
+    "tags": [
+        "UI related",
+        "prompting"
+    ]
+}


w-e-w:
your extension works but you need to be careful about `GradioDeprecationWarning`
```py
creaprompt.py:67: GradioDeprecationWarning: The `style` method is deprecated. Please set these arguments in the constructor instead.
  Prefix = gr.Textbox(label="Prefix of the Prompt", elem_id="promptgen_prompt_prefix", show_label=True, lines=2, placeholder="Type your prefix or leave blank if you don't want it").style(container=True)
```
also your extension seems to provide in duplicate functionality to the existing ones such as [wildcards](https://github.com/AUTOMATIC1111/stable-diffusion-webui-wildcards) with integration [tagcomplete](https://github.com/DominikDoom/a1111-sd-webui-tagcomplete) extension? or other prompt symptoms extensions such as [dynamic prompt](https://github.com/w-e-w/sd-dynamic-prompts)
is there any reason why one would use your extension over the existing ones that have integration with other extensions?

w-e-w:
so I took your CSVs and rename them to txt add them to wildcard dir of [wildcards](https://github.com/AUTOMATIC1111/stable-diffusion-webui-wildcards) and use [tagcomplete](https://github.com/DominikDoom/a1111-sd-webui-tagcomplete) to call them

image

[wildcards](https://github.com/AUTOMATIC1111/stable-diffusion-webui-wildcards)
- [x] flexibility to choose, what will be included in the result
- [x] can easily include our own categories
- [x] and the order in which they will appear in the result
- [x] You can add your category easily 

from what I can see this is the same function of you as your extensions

---

maybe I'm missing something obvious
if you're sure that this is useful then I will add it into index

if you continue to develop I suggest you stop calling your categories CSVs because there are not, CSV is [Comma-separated_values](https://en.wikipedia.org/wiki/Comma-separated_values) you're using New Line separated values, which is 
exactly the same format used by [wildcards](https://github.com/AUTOMATIC1111/stable-diffusion-webui-wildcards) they're just using txt extension
you might as well make it compatible with [wildcards](https://github.com/AUTOMATIC1111/stable-diffusion-webui-wildcards) ext and call you categories wildcards instead as it is already a turn that is being used around

tritant:
Thanks for your suggestions, I'll look into it. I updated the script to remove the gradio warnings.

tritant:
Hello, do you intend to do the merge or not ? Because otherwise I will broadcast the github url.

w-e-w:
oh I was waiting on you to fix the gradio warning
I thought you were going to message me when you finished

PR #291, opened 2024-03-23
Update and rename sd-webui-tinycards.json to sd-webui-cardmaster.json

SenshiSentou:
## Info 
Kept the git URL the same as it redirects. This will (hopefully?) allow users to simply update their existing Tiny Cards extension. New URL is https://github.com/SenshiSentou/sd-webui-cardmaster

## Checklist:

### DIFF
--- a/extensions/sd-webui-cardmaster.json
+++ b/extensions/sd-webui-cardmaster.json
@@ -1,7 +1,7 @@
 {
     "name": "Card Master",
-    "url": "https://github.com/SenshiSentou/sd-webui-tinycards.git",
-    "description": "Adds powerful tools to the \"extra network\" cards (esp. LoRAs), including detail views, easy application of activation texts to the negative prompt, partial application of activation texts that contain multiple sections, and automatic prompt cleanup when removing a LoRA.",
+    "url": "https://github.com/SenshiSentou/sd-webui-cardmaster.git",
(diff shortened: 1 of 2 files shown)


w-e-w:
keeping or changing the URL on the index is irrelevant to if the user is able to update the extension
as updating is independent from the index
extension update just a git pull as long as GitHub still redirects the old URL it will be able to update

---

changeing the index will only have an effect on new installs of the extension 
when installing it will be install undere the dir

if the URL is not change
new installes will be install under the old name `sd-webui-tinycards` and not the new name `sd-webui-cardmaster`
and this is the name they will see when the on the `extensions tab` for enable / disable extensions

if the URL is change
the it will be install under the new name `sd-webui-cardmaster`
and will be easyer to find under the `extensions tab`
the only donwside is that if a user has install `sd-webui-tinycards` in the past if they go to the `available extensions tab` since it's under a diffrent name, webui will thinks it has not been installed, and show the install button and let the user install it
which might cause the extension to be installed twice and possibly leads to issues

---

I'm not sure what your intention or for changing URL but if your intention is to let the user recognize this under a different name
then I suggest you change the url as well
if you don't care about the install name directory and how it's showned on the tab then not changing the URL is fine

---

either way I will suggest you add the old name `tiny cards` somewhere in the description
this way that when an old user who knows the extension by the old name and search for it will be easier to find when they wish to re-install the extension

SenshiSentou:
Gotcha, I didn't realize that. Changed the URL and mentioned the name change in the description. I also added a check with warning popup in case the old version is still installed

Screenshot 2024-03-24 at 13 15 58

PR #299, opened 2024-04-04
Added Batchlinks Downloader

BarrenWardo:
## Info 

[Batchlink Downloader](https://github.com/etherealxx/batchlinks-webui.git)

An awesome extension missing for a long time now.

## Checklist:

> Apologies if there are any mistakes.

### DIFF
--- a/extensions/batchlinks-webui.json
+++ b/extensions/batchlinks-webui.json
@@ -1,0 +1,10 @@
+{
+    "name": "Batchlinks Downloader",
+    "url": "https://github.com/etherealxx/batchlinks-webui.git",
+    "description": "Download several Huggingface, MEGA, and CivitAI links at once. SD webui extension. For colab.",
+    "tags": [
+        "script",
+        "tab",
+        "online"
+    ]
+}


w-e-w:
sorry I meant to test and reply earlier but one thing that to another

there is quite a big issue with this extension

first from wanting to tell it seems it's no longer being maintained
this itself is not a death sentence but I didn't do this there's a big issue about this extension
which is it does not actually download the models to the correct path that webui uses
it only downloads to them to the default dir
but the default path can be change if the uses uses [Command Line Arguments and Settings](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Command-Line-Arguments-and-Settings) like `--data-dir` / `--ckpt-dir` 

I consider one of the basics requirements of model downloader is that it's able to download the models to the correct directories

if I want to I can fix this issue relatively easily, the problem is even if I do submit a PR to fix this it's not going to be merged
that's the maintainer of index I could create a fork and point the index to a fixed fork
but I personally won't have the energy to maintain this extension, someone else would have to do it

the conclusion
from my perspective unless someone is willing to fix the current issues and maintain the extension I don't want to add this to index

BarrenWardo:
@w-e-w Sorry for the delay in reply. Somehow I missed the notification.

First of all, I do agree that this extension uses default location by default but if you check the documentation. It has option to create custom hashtags for custom locations [here](https://github.com/etherealxx/batchlinks-webui?tab=readme-ov-file#custom-hashtag-path-new). 

I agree that the dev is not maintaining the repo anymore & would be really good if he starts maintaining again or someone else takeover this. But even now it is fully functional & efficient extension compared to many other downloading extensions.

w-e-w:
WTF
[`Running Shell Commands`](https://github.com/etherealxx/batchlinks-webui?tab=readme-ov-file#running-shell-commands) as a "future" without any safeguard even when webui is accessible remotely
(like when `--listen` is used)
I hope the developers of do understand that that is a colossal security exploit
what are they even thinking just enable it like this without safeguards

server side arbitrary location any file download alone is also a exploit
as certain files in certain locations can be automatically executed
there's such dir inside webUI and there's such places inside the OS

but `arbitrary code execution` is on another level 

there was another extension that adds arbitrary code execution
did the developers knows about the risk of this and tries to inform users I helped him add a couple of extra safeguards and then added to the index
- https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/278#issuecomment-1937662292

---

I have to adjust my assessment
- has to download models to the correct locations
- configuration of download location should be put behind safeguards to prevent arbitrary location file download
- `arbitrary code execution` and possibly other major exploits needs to be put also behind safeguards
- they should be automatically disabled when webui is accessible from the internet
- only after the user has fully understand the implications can those features can it be enabled
- enabling of the settings must be done in a secure fashion

otherwise this succession cannot be listed on index

PR #307, opened 2024-04-25
add readme browser

light-and-ray:
## Info 
https://github.com/light-and-ray/sd-webui-readme-browser
This extension allows user to view readme files of extensions locally right inside sd-webui

## Checklist:

### DIFF
--- a/extensions/sd-webui-readme-browser.json
+++ b/extensions/sd-webui-readme-browser.json
@@ -1,0 +1,9 @@
+{
+    "name": "Readme browser",
+    "url": "https://github.com/light-and-ray/sd-webui-readme-browser.git",
+    "description": "This extension allows user to view readme files of extensions locally right inside sd-webui",
+    "tags": [
+        "dropdown",
+        "UI related"
+    ]
+}
--- a/extensions/sd-webui-readme-browser.json
+++ b/extensions/sd-webui-readme-browser.json
@@ -1,9 +1,10 @@
 {
     "name": "Readme browser",
     "url": "https://github.com/light-and-ray/sd-webui-readme-browser.git",
-    "description": "This extension allows user to view readme files of extensions locally right inside sd-webui",
+    "description": "This extension allows user to view readme files of extensions locally right inside sd-webui. Online will be used only if readme contains external media and links",
     "tags": [
         "dropdown",
+        "online",
         "UI related"
     ]
 }


w-e-w:
tested extension works but I have some concerns

my main concern is external assets
as it will access external resources if it's in readme
so the `online` tag is required

but even with the `online` tag I think it's a good idea to have an option to not load external resources by default
> similar to email services
images hosted on github is "probably" fine, but any other assets should probably be guarded behind a checkbox or button

also links
I think they should be a pop up warning users they're about to leave web UI I'm clicking on any external links

theres also the potential possibility of people embedding malicious JavaScript once they know such an extension like this exist
> then again if some one wish to do harm it's much easier to directly write python code so this isn't really that big of a deal...

what do you think

---

> btw I don't think you know this, I actually had the same exact idea but just didn't end up completing it, I called my version 'docs'
> my intention at the time was to first implement it as an extension and depends on the reception maybe implemented as a built-in extension, that's probably why I raise those concerns above

w-e-w:
> > so the online tag is required
> 
> Can be added. But the main feature is give access even if the user offline and this flag can confuse people But the description exists so no problem

if some is offline as in disconnected from the internet
then they're most likely using web UI locally (they technically can be on a lan network running webui on another server)
if using locally they might and need acaccess to the readme, they can open the readme directly on there PC without the extension
and so they made advantage of extension is not offline access but more ease of use

> > but even with the online tag I think it's a good idea to have an option to not load external resources by default
> 
> I don't think so. Why? They don't sent any data e.g. prompts, models etc. Only ip and referer if the server is not local. Also it's hard to implement

but the reality is lots of people put images on GitHub
which means those pages won't fully function offline
from the couple pages I've test I already found lots of examples
the most common ones connect to GitHub, but some connects to Buy Me a Coffee
some of these external server can provide tracking functionality, the most benevolent is a view counter
but in worst case they could be using gradio share and the browser decides to add a referral link in the request payload including the full URL of that share instance so someone can potentially access the link and gain access to their instance
> I'm not entirely sure how this happens but I something like this happened in the past on GitHub repo traffic

some people wish to not connect to anything at all, I've seen people complaining about webui access Google fonts and stuff

> tbh I think the majority of people that don't care about connecting to external service would probably just click the extension URL that takes them to get up on the extensions tab
> and the ones who are very sensitive about this probably have extra measures to prevent internet access

but the urrent `online` tag rule is essentially
if a extension needs online access to function, aside from a few exceptions such as one time downloading of assets like modles not consequential of the checks to GitHub (generally regarded as "safe") or using PIP during install, then it needs the online tag
in this case there is a chance of it requesting images from external source

in this case the extension may accessing external resources like images outside of users control during normal use, so it requires the online tag

```
it is not about if extension can function offline, it's more about if the extension will access online
```

---

> > images hosted on github is "probably" fine
> 
> Not fine, why people can't put them in repo XD I've made caching of them. Also have made for imgur, but it gives me Http 429 error 😵‍💫 I want to decrease numer of requests but they do the vice versa

miscommunication what I meant by "fine" is from a security sensitivity standpoint
since webui user are basically already using GitHub, and as far as I know GitHub does not allow access of the logs on who are accessing which files

---

> > Also I think it's a very bed idea to add in webui directly because browsing readme and doc files for fun doesn't fit to it completly

you miss the next sentence
> "depends" on the reception maybe implemented as a built-in extension

PR #308, opened 2024-04-25
Add prompt-from-text-or-tagger, expanded-xyz-plot, clipboard extensions

GiusTex:
#### Links:
- [prompt-from-text-or-tagger](https://github.com/GiusTex/sd-webui-prompt-from-text-or-tagger). Like the prompt-from-file-or-textbox script, but with built-in grid-maker and auto-tagger option. Useful to make random images and test prompts based on other images, automatically.
- [expanded-xyz-plot](https://github.com/GiusTex/sd-webui-expanded-xyz-plot). Updated version of [yspacedevyspacedev Expanded-XY-grid](https://github.com/yspacedev/Expanded-XY-grid). Adds more features to the standard xyz grid, such as putting multiple parameters in an axis row. Useful if you want to test more than 3 parameters at a time.
- [sd-webui-clipboard](https://github.com/GiusTex/sd-webui-clipboard). Extension that adds a textfield where to store notes, prompt parts, parameters etc between sessions. Useful to store prompt parts and parameters while testing.

I didn't know if it was better to make 3 separate requests or not (in case some didn't go well and prevented the others from being loaded), I hope that a single one is right.

The extensions are maintained.

## Checklist:

### DIFF
--- a/extensions/sd-webui-clipboard.json
+++ b/extensions/sd-webui-clipboard.json
@@ -1,0 +1,8 @@
+{
+    "name": "sd-webui-clipboard",
+    "url": "https://github.com/GiusTex/sd-webui-clipboard.git",
+    "description": "Extension that adds a textfield where to store notes, prompt parts, parameters etc between sessions. Useful if you want to store prompt parts and parameters while testing",
+    "tags": [
+        "script"
+    ]
+}
--- a/extensions/sd-webui-expanded-xyz-plot.json
+++ b/extensions/sd-webui-expanded-xyz-plot.json
@@ -1,0 +1,8 @@
+{
+    "name": "sd-webui-expanded-xyz-plot",
+    "url": "https://github.com/GiusTex/sd-webui-expanded-xyz-plot.git",
+    "description": "Updated version of yspacedevyspacedev Expanded-XY-grid. Adds more features to the standard xyz grid, such as putting multiple parameters in an axis row. Useful if you want to test more than 3 parameters at a time",
+    "tags": [
+        "script"
+    ]
+}
--- a/extensions/sd-webui-prompt-from-text-or-tagger.json
+++ b/extensions/sd-webui-prompt-from-text-or-tagger.json
@@ -1,0 +1,11 @@
+{
+    "name": "sd-webui-prompt-from-text-or-tagger",
+    "url": "https://github.com/GiusTex/sd-webui-prompt-from-text-or-tagger.git",
+    "description": "Like the prompt-from-file-or-textbox script, but with built-in grid-maker and auto-tagger option. Useful to make random images and test prompts based on other images, automatically",
+    "tags": [
+        "script",
+        "prompting",
+        "editing",
+        "query"
+    ]
+}


w-e-w:
I haven't actually test the extensions yet need (to go to bed)

but one quick comment about expanded-xyz-plot
> This script is very heavily based (I just modified the code) on the original xy_grid.py created by AUTOMATIC1111 for stable-diffusion-webui, so it includes all the same features. And thank you AUTOMATIC1111 and all other contributors for the original script.

since this is the case I believe AGPL LICENSE is required https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/LICENSE.txt

GiusTex:
Thanks for the info, I'll update the extension page. Just to be sure, I just have to add the license.txt? And maybe credit auto1111

w-e-w:
an extension should be in working order for it to be listed on index 
they should be relatively bug free and reach a level of polishness
in order to make sure that this is case I actually do code review and test the extension
and after conducting my review I came to conclusion that you did not actually test your own extension

you submitted three extensions
all three of them have different levels of issues

---

[prompt-from-text-or-tagger](https://github.com/GiusTex/sd-webui-prompt-from-text-or-tagger)
is utterly broken
not only that you in a specifically decide to hardcode the model used for wd14 it's not even a default model as far as I can tell
your great images safe directory is also invalid, first you decide to hardcore the default web you are out the path, but you did not realize that those paths may not exist so you need to create the ooutput dir otherwise your script will error, you should also access the output directory set by the user `hared.opts.outdir_img2img_samples` and not hardcode them, or better create your own setting field you know how to do that in a different extension
you also should not use the API to call wd14 in the first place, using API when it's not necessary it's just bad, Implement your own function to call wd14 internally on server side without using API, have access to the list of models that the user have and that the user choose which model to use as opposed to hardcode

---

[expanded-xyz-plot](https://github.com/GiusTex/sd-webui-expanded-xyz-plot)
`prompt matrix` just doesn't work
and the reason is you don't have an understanding of either the original XYZ script or https://github.com/yspacedev/Expanded-XY-grid, you only take sections of one script and paste it into the other
that doesn't work
not only that because of this script is implemented as a complete standalone script apart from the original XYZ script
the usefulness of `Multitool` is largely handicapped
the thing about the origin XYZ script is that it is commonly extended by different extensions
for example  the built-in `Hypertile` or third party `ControlNet` 
these three additional axes added by `expanded-xyz-plot` should not be implemented as a set of script instead it should be implemented as an addition to the original XYZ script
I'm assuming that the original creator either did not know that this was possible or it wasn't actually possible back then
how do I know this is possible now, because I decided to take matters into my own hands and implement it myself
my version just add extra axis into the built-in xyz script which means these axes can be used with any other axes added by extensions or future update

image

> after I have finalized my implementation of documentations I will be listed them extension on index

---

[sd-webui-clipboard](https://github.com/GiusTex/sd-webui-clipboard)
this is the only extension that is technically working without major issues
but it is a very very bare bones extension and some of the code just don't make sense
why the heck does changing the directory need to restarting
having the user greeted with an error message if they decide to click load before they click save is not exactly a good look for such a simple extension
lots of strange code that serves no purpose

and again I decided to take matter into my own hands
and just making extension with similar purpose but far more polished
https://github.com/w-e-w/sd-webui-notepad
the main difference between my implementation is that my implements in actual history and allows creation of unlimited number of text if the user wishes and even multi-user support when `--gradio-auth` is used

image

> again after I have finalized my implementation of documentations I will be listed them extension on index

---

Conclusion

all three of the extensions have issues ranging from minor to major listed below

- [prompt-from-text-or-tagger](https://github.com/GiusTex/sd-webui-prompt-from-text-or-tagger)
doesn't work fix
it needs to be fixed in order for it to be listed on index,
best implemented directly via python without API, but if API must be used it must be seamless to the user, "seamless" as in the user does not need to manually input the IP address (credentials when authentication is required), they should be a method to select the model
the output pass should be correct and possibly other issues

- [expanded-xyz-plot](https://github.com/GiusTex/sd-webui-expanded-xyz-plot)
declined, partially doesn't work in implemented in the not very useful manner,
I have implemented alternative as far as can tell it's obsolete now

- [sd-webui-clipboard](https://github.com/GiusTex/sd-webui-clipboard)
while this extension is technically working, I feel like it's very Bare Bones and is still in working progress extension
since I have made an extension that I considered to be a far more polish or round
in order for your extension to be listed it needs to be in a better State and offering more than my extension
you can have it resubmitted

---

after review all 3 of your extentions, I'm not confident in your ability to create and maintain these extensions
and so I've decided to decline your entrys

you have the option to resubmit your extension in the future

if you're just writing extensions for yourself then no one cares about what you do
but when you submit an extension to index you're basically saying you wish to share your extension with the world and asking our endorsement you must first test your extensions extensively

GiusTex:
Thanks for the precious directions, I admit that I am not a programmer and that I created these extensions for myself. Maybe if I had created these extensions with the idea of releasing them to a wider audience they would have taken different paths (I would have known that not everyone would have my directories, variables etc). Maybe the other github projects I made made me overconfident and made me think to share them with just few changes (the ones that came to mind when faced with the amount of code, and I also dind't have more ideas on how to test my scripts at the end).
Below I justify myself for the errors, some due, some avoidable, etc., anyway if you want you can skip them. You will probably see some noob thoughts.

<details>
<summary>Thoughts on the errors</summary>

[prompt-from-text-or-tagger](https://github.com/GiusTex/sd-webui-prompt-from-text-or-tagger)
> you hardcode the model used for wd14, the model it's not even a default model

You're right about the model, I forgot I had set a more recent one; the other models made available by default by the wd14-tagger extension can still be used (I thought about a dropdown/textbox menu but didn't thought of adding it).

> your grid images save directory is also invalid

I admit I didn't thought other uses can have different directories. This extension probably is better on my computer, as it is...
> you also did not make sure those path exist

It probably has to do with the message above and the use of save settings (if one doesn't save images in folders divided by date, you're right)
> you should use the user setting `hared.opts.outdir_img2img_samples`

I didn't know of that option, my fault here.

> implement your own method to call wd14 internally on server side without using API

I tried that, but truly didn't know how to call the extension. I searched online, in other extensions, on github, but I'm not expert and landed on this inefficient way.

[expanded-xyz-plot](https://github.com/GiusTex/sd-webui-expanded-xyz-plot)
> `prompt matrix` just doesn't work

I admit that I have never used 2 of the 3 tools added by the original author, I was only interested in putting more options in the axes, and out of inertia I didn't remove the rest of the code. Turning it into a public extension I should have removed the leftover two tools probably. Thanks for the [w-e-w/sd-webui-xyz-addon](https://github.com/w-e-w/sd-webui-xyz-addon); I think you should add some info on how to use the new options though: I tried `Steps: 10, 20 | Sampler: Euler a, DPM++ SDE Karras` and got two times `'NoneType' object has no attribute 'options'`, probably I should use a differnt sintax but I don't know what it is (I still got the grid, with half the images)

> it also quite common to have new parameters / axis added to it

I agree, they are very nice, but if I remember well maybe their code was too big for an inexperienced programmer like me to understand which parts added the options, and which made them work, so I continued with that old extension I had "understood".

[sd-webui-clipboard](https://github.com/GiusTex/sd-webui-clipboard)
> why the heck does changing the directory need to restarting

You need to restart for the changes to take effect (until you restart the files are stored in the old directory)

> lots of strange code that serves no purpose

I changed the code from [example-send-text](https://github.com/EnsignMK/ExampleSendText), I thought I removed the unnecessary parts, I must have missed some.
And thanks again for [w-e-w/sd-webui-notepad](https://github.com/w-e-w/sd-webui-notepad), it looks much better.

If you could credit me at least for the idea behind it (the notepad script) I would appreciate.

> you must first test your extensions extensively

I'm fine with fixing errors, I just dind't have more ideas on how to test my scripts, and thought I would have fixed the ones bringed by the users.

</details>

Thank again for your thoughts and the updated scripts of yours.

w-e-w:
I made a typo
`shared.opts.outdir_img2img_samples` not `shared.opts.outdir_img2img_samples`
this is basically how you access every setting inside web UI
and you should know this because you have tried to access you own setting value `shared.opts.data.get("clipboard_path", os.path.join(script_dir, "scripts")` except using an overkill method

and this is also related to
>>why the heck does changing the directory need to restarting
>You need to restart for the changes to take effect (until you restart the files are stored in the old

what I ment was there is no reason why change a setting change will need restarting to update

---

no I keep the syntax the same and improved on it documentation will come later when I have time
the reason why failed
```
Steps: 10, 20 | Sampler: Euler a, DPM++ SDE Karras
'NoneType' object has no attribute 'options'
```
has nothing to do with the extension
in webui 1.9 Samplers and scheduler are separated
`DPM++ SDE` `Karras`
and so `DPM++ SDE Karras` it's not a valid option
```
Steps: 10, 20 | Sampler: Euler a, DPM++ SDE
```
image

w-e-w:
> Below I justify myself for the errors, some due, some avoidable, etc., anyway if you want you can skip them. You will probably see some noob thoughts.

being a noob fine we all are noob once
and the more you learn the more things you realize you don't know
you do not have to delete your extensions I wasn't asking of you to do that
I spent the time typing up your issues in hoping that you can improve on them not delete them

GiusTex:
> this is basically how you access every setting inside web UI and you should know this

Thanks for the explanation; I didn't know it, I adapted it from the past-text-example extension I think, and thought it had a more limited use.

> I wasn't asking of you to do that

I knew, I thought though that since you made your own versions, mine would be unecessary, since you also said `after I have finalized my implementation of documentations I will be listed them extension on index` (under your sd-webui-xyz-addon image, in the quotation).

> I spent the time typing up your issues in hoping that you can improve on them

Sorry, I couldn't have known that. Surely it will be useful in the future, but I think your 2 script are better than my versions. I could maybe update the prompt-from-script-or-tagger (I still have everything locally), I just thought that fixing all the errors would be a nuisance. You smashed my dreams a bit, before

w-e-w:
> (how many people would actually need my script? Not many, I think)

"need" and "will try it" are totally different
I think you be quite amazed on the amount of people that are willing to give a extension a try
but for them to even "try", it needs to work

I for some strange reason have a tendency of writing extensions that I don't use myself
I even wrote an extension that I recommend people not to use
if I recall correctly initially lose hundreds of people per day and now after almost a year they're still like 20 people cloneing it everyday

this is one of the reasons why I insist on extension to be in a good enough state with it initially is added to the index
lots of people are going to try it, and it will looks very bad if it's "Dead on Arrival"

GiusTex:
I'll change the prompt-from-text-or-tagger then, with calm

GiusTex:
I didn't make that forked extension, GuinnesShep did

w-e-w:
ooooooooh very sorry
lol my bad
you two have similar color profile picture
and both start with the letter G

---

anyway I'm busy recently
sorry for the delay about the two extensions

my version of the XYZ multi-tool should be usable and rid of most bugs with a couple of minor issus that needs fixing
should be able to release it soon

as for the notepad / clipobard, I'm trying to get a JavaScript pop-up thinkg working
once that is done it should be good for release

GiusTex:
> you two have similar color profile picture and both start with the letter G

Ahah ok :)

> sorry for the delay about the two extensions

No problem, I already find it remarkable that someone has decided to take charge of these little extensions of mine

PR #311, opened 2024-05-08
New Extension SD-M2M-Animate

chalcids:
## Info 
[Repo URL](https://github.com/chalcids/sd-m2m-animate)

## Checklist:

### DIFF
--- a/extensions/sd-m2m-animate.json
+++ b/extensions/sd-m2m-animate.json
@@ -1,0 +1,9 @@
+{
+    "name": "SD-M2M-Animate",
+    "url": "https://github.com/chalcids/sd-m2m-animate.git",
+    "description": "Allows users to create videos from a base, generate test frames to ensure correct parameters, upscale videos to desired resolutions, while using SD and controlnet as the backbone. This is compatible with webui 1.8",
+    "tags": [
+        "tab",
+        "animation"
+    ]
+}


w-e-w:
extensions need to work with the current version of webui at the time of submission in order to be listed 
I believe you're coded your extension based on 1.8
but it's broken in 1.9 due to the sampler scheduler rework

> note the following test is done on 1.8 since it doesn't work on 1.9

you also have stuff in your car that can potentially break stuff
- `Reload UI` is broken (webupage footer)
- javascript `function setRandomSeed`, you need to remember that the JavaScript namespace is shared globally, which means dysfunction can override or be override, there's no point re-defining the same function that webui already has, since it's the same function it's not causing issues now but if in the future this function is changed on the webui, the existence of you redefining the function would either cause your function to be overwritten or overwritten the function nearly defined by web UI which causes things to break
- also as a general rule you should not use generic names in JavaScript otherwise you could run into the risk of name collisions
prefix or suffix it with something like your `submit_m2m_animate`
- you patch gradio fix id, I'm assuming that you did this because you kind of want to allow usage of other extensions under your tab and you want to avoid the issue of having two elements with the same ID because when this happens gradio will randomly generate the ID, this should be fine as long as your extension tab is created last but that may not necessarily be the case, if I recall correctly setting `UI tab order` can change the tab order and cause you would have to be loaded first which would cause the elements on your tab to have the origin ID, I don't really have a good solution for this issue
maybe instead of creating your own tab usually create a operation mode under image image like
https://github.com/John-WL/sd-webui-inpaint-background ,he built a pip package `sdwi2iextender` to "help create custom img2img operation modes in A1111". https://github.com/John-WL/sdwi2iextender
or implement this as a drop-down script (scripts like xyz grid)
to avoid the issue of creating your own tab and all ui are interfaces
- tool buttons except clear promps don't work, reconnect button is shown when it's supposed to be hidden, style saving doesn't work I suggest you remove the ones that don't work
image

--- 

- generally speaking you should only have one entry file all other things should be in all of directories so as to not have some strange import tree which might cause strange issues that's all hard to debug
for example of a safer structure see my extension
https://github.com/w-e-w/sd-webui-hires-fix-tweaks
theres only one script under scripts, all the things are under different directory, whiles there are cases to you may want to have multiple files under scripts but it;s best not too

PR #312, opened 2024-05-08
changing to fork with more up-to-date scripts

gutris1:
[Repo](https://github.com/gutris1/sd-fast-pnginfo)

### DIFF
--- a/extensions/sd-fast-pnginfo.json
+++ b/extensions/sd-fast-pnginfo.json
@@ -1,7 +1,7 @@
 {
     "name": "Fast PNG Info",
-    "url": "https://github.com/NoCrypt/sd-fast-pnginfo.git",
-    "description": "Allows you to view PNG Info of an image without upload to webui's backend server, significantly faster when webui is hosted remotely like on Colab.",
+    "url": "https://github.com/gutris1/sd-fast-pnginfo.git",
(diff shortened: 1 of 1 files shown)


w-e-w:
@NoCrypt
- https://github.com/NoCrypt/sd-fast-pnginfo/pull/9

CapsAdmin:
> Hmm, does curl program exist on windows? In install.py it is used. Maybe it's better to use `urllib.request.urlretrieve(url, outPath)` for compatibility

https://curl.se/windows/microsoft.html

Yeah, since around windows 10. Whether or not that means you should use it is something else. I'd say if you're in python already, just use python.

gutris1:
oops I didn't even think about Windows, fixed now
thanks guys

gutris1:
closing this as I've lost interest.

gutris1:
> * left a message at the wrong place so reposting here again [UPDATE NoCrypt/sd-fast-pnginfo#9 (comment)](https://github.com/NoCrypt/sd-fast-pnginfo/pull/9#issuecomment-2103915328)
> 
> just so you know when possible we don't like to change the url of a extension on the index the reason being changing the URL only changes it for new installations and based on NoCrypt account activity he is still active just last week so in this case it's best to be contact the original repository
> 
> NoCrypt just replyed [NoCrypt/sd-fast-pnginfo#9 (comment)](https://github.com/NoCrypt/sd-fast-pnginfo/pull/9#issuecomment-2103921649)

I understand that, that's why Im closing this PR
Thank you.

PR #319, opened 2024-05-24
m9 prompt extensions

MarcusNyne:
## Info 
There are two extensions included here.  They have similar functionality but have key differences.  I have been using them extensively and found them to be very useful.  There is detailed documentation in README.md.  Please review for more details.  The direct github urls are:

https://github.com/MarcusNyne/sd-scramble-prompts-m9
https://github.com/MarcusNyne/sd-tweak-weights-m9

Feel free to discord me any questions. @marcusnyne

Thank you!!!

## Checklist:

### DIFF
--- a/extensions/m9-scramble-prompts.json
+++ b/extensions/m9-scramble-prompts.json
@@ -1,6 +1,6 @@
 {
     "name": "ScramblePrompts [m9]",
-    "url": "https://github.com/MarcusNyne/sd-scramble-prompts-m9",
+    "url": "https://github.com/MarcusNyne/sd-scramble-prompts-m9.git",
     "description": "Creates prompt variations during image generation by reordering prompts, removing prompts, and modifying prompt weights.  Helps to fine-tune your prompts.  Details in README.md",
     "tags": [
         "prompting"
--- a/extensions/m9-scramble-prompts.json
+++ b/extensions/m9-scramble-prompts.json
@@ -2,7 +2,6 @@
     "name": "ScramblePrompts [m9]",
     "url": "https://github.com/MarcusNyne/sd-scramble-prompts-m9",
     "description": "Creates prompt variations during image generation by reordering prompts, removing prompts, and modifying prompt weights.  Helps to fine-tune your prompts.  Details in README.md",
-    "added": "2024-05-24",
     "tags": [
         "prompting"
     ]
--- a/extensions/m9-scramble-prompts.json
+++ b/extensions/m9-scramble-prompts.json
@@ -1,0 +1,9 @@
+{
+    "name": "ScramblePrompts [m9]",
+    "url": "https://github.com/MarcusNyne/sd-scramble-prompts-m9",
+    "description": "Creates prompt variations during image generation by reordering prompts, removing prompts, and modifying prompt weights.  Helps to fine-tune your prompts.  Details in README.md",
+    "added": "2024-05-24",
+    "tags": [
+        "prompting"
+    ]
+}
--- a/extensions/m9-tweak-weights.json
+++ b/extensions/m9-tweak-weights.json
@@ -1,0 +1,9 @@
+{
+    "name": "TweakWeights [m9]",
+    "url": "https://github.com/MarcusNyne/sd-tweak-weights-m9",
+    "description": "Creates prompt variations during image generation by modifying prompt weights matching specified keywords.  Helps to fine-tune your prompts.  Details in README.md",
+    "added": "2024-05-24",
+    "tags": [
+        "prompting"
+    ]
+}
--- a/extensions/m9-tweak-weights.json
+++ b/extensions/m9-tweak-weights.json
@@ -1,6 +1,6 @@
 {
     "name": "TweakWeights [m9]",
-    "url": "https://github.com/MarcusNyne/sd-tweak-weights-m9",
+    "url": "https://github.com/MarcusNyne/sd-tweak-weights-m9.git",
     "description": "Creates prompt variations during image generation by modifying prompt weights matching specified keywords.  Helps to fine-tune your prompts.  Details in README.md",
     "tags": [
         "prompting"
--- a/extensions/m9-tweak-weights.json
+++ b/extensions/m9-tweak-weights.json
@@ -2,7 +2,6 @@
     "name": "TweakWeights [m9]",
     "url": "https://github.com/MarcusNyne/sd-tweak-weights-m9",
     "description": "Creates prompt variations during image generation by modifying prompt weights matching specified keywords.  Helps to fine-tune your prompts.  Details in README.md",
-    "added": "2024-05-24",
     "tags": [
         "prompting"
     ]


w-e-w:
- to be honest I would prefer you to merge these two extension into one as both of these extensions are almost copies of each other sorry from the generation code
you can create a setting that allows the user to enable or disable the two modules separately. you could also create an extension with sub multiple tabs

### issues that are in both extensions

----

- you creating a sub Loop of P inside the pipeline makes me uncomfortable, as far as I'm aware this can break things even though I don't have a concrete example, both of your scripts are similar in nature to XYZ-grid script "a selectable script", the main difference between selectable scripts and a always on script is that a selectable script has full control of the pipeline. they can copy p as a wish without fear of any issues, it is quite possible some extensions might have some operations that span across multiple callbacks and would need to be reseted by the initialization of P or prior callbacks. creating a sub Loop of P which does not have the exact same initialization sequences could have potential issues

---

- the output dir of images is not just `p.outpath_samples` sub dir `shared.opts.directories_filename_pattern` setting also needs to accounted for
- 

`Create info text file`
this means getting the output directory name is not as straightforward
I believe it may simplify the `writing of info` if you register a call back with [`on_before_image_saved` `on_before_image_saved`](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/1c0a0c4c26f78c32095ebc7f8af82f5c04fca8c0/modules/script_callbacks.py#L499-L512)
this way you directly have access to the final file path where the image is saved

`Create variation folders`
if you want to create a subdirectory group it is also possible for you to "temporarily" append to the following pattern of
`shared.opts.directories_filename_pattern` and unabe `save in sube dir` then restroing the setting beck
> note as the name would suggest `directories_filename_pattern` is a pattern similer to the `image file name pattern`

code like this shoud work
```py
# save original settings 
original_directories_filename_pattern = shared.opts.directories_filename_pattern
original_save_to_dirs = shared.opts.save_to_dirs

if chk_variation_folders:
    original_directories_filename_pattern
    try:
        if shared.opts.save_to_dirs:
            shared.opts.directories_filename_pattern = shared.opts.directories_filename_pattern + '/[datetime<%Y%m%d-%H%M%S>]'
        else:
            shared.opts.directories_filename_pattern = '[datetime<%Y%m%d-%H%M%S>]'
        shared.opts.save_to_dirs = Ture
        
        # image save after this before will be saed into sub dir
        
    finally:
        # restore original setting in final block to make sure the setting is restored
        shared.opts.directories_filename_pattern = original_directories_filename_pattern
        shared.opts.save_to_dirs = original_save_to_dirs
```
> this is hand to type untested code

---

- the overcomplicated JavaScript implementation of module enable indication is partially broken
> it dose not worrk on img2img tab
image
- please just use [InputAccordion](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/1c0a0c4c26f78c32095ebc7f8af82f5c04fca8c0/modules/ui_components.py#L88-L92) to indicate that your extension is enabled, an input accordion is a custom gradio element of webui, is a checkbox combined with an accordion, this is the type of element that is used by `hires fix` a refine and other extentions
- you basically made the same mistake as another extension author https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/315#issuecomment-2109846694 read section `issue UI JavaScript and elem_id` [extra additional conversation](https://github.com/muerrilla/sd-webui-detail-daemon/pull/1)
I can tell that you have adopted indication from `control-net`, but their use case is different from your, they requires indication of multiple sub-unit, where as your extention only requires one

using `InputAccordions` removed all the JavaScript (apart from the hints) and CSS  making a code base easy to maintain

image

> if you go with the sub multiple tabs route I mentiond above then I have no issue with you using controlnet style 

---

your code structure can lead to potential issues in the future

best restructure it like so
```
extention root
├── README.md
├── your_libs_under_unique_name_across_webui/
│   ├── __init__.py                           <-- optional: depends on if you want to make a package or not doesn't really matter
│   ├── your_lib_1.py
│   └── your_lib_2.py
└── scripts/
    └── best_to_have_one_file_only.py
```
in best_to_have_one_file_only.py
```py
import your_libs_under_unique_name_across_webui
```

---

### issue with `Tweak Weights`
it't doesn't seem to function correctly

I'm not sure what your intention is either it's supposed to replace prompts or replace lora
for me when it replaces prompts with different weights it literally just appended `:0.123` after the prompt, but without the parenthesis so the syntax is incorrect, you also have a lora weight, but I've also wasn't able to get it working with lora so that is extra confusing

---

conclusion

- I think the method your implementing extension is a bit unorthodox and can have potential issues, but as it seems to function I will let this one go but please watch out for potential compatibility issues

- unless you vehemently object I will ask you to combine your extensions

- fix enable indication issues
> preferably using InputAccorden the pre-existing solution that is known to work, as opposed to reinventing the wheel

- fix the output directory path to respect user subfolder and `directories_filename_pattern` settings

- remove tha `"added": "2024-05-24",` from the entry, it will be automatically generated on merge

---

transparency I'm currently also working on an extension, but it's purely function as an extension of XYZ
I will say your use case is a bit different from mine, but there's also quite a crossover

w-e-w:
what does `[m9]` mean?

MarcusNyne:
@w-e-w , thank you for taking the time to offer comprehensive feedback.  Somehow I didn't notice your comment, but I noticed when @light-and-ray commented.  Many of your comments are highly technical and I will need to review in detail.  I will offer some initial responses, though.

* I wasn't able to find documentation for implementing complex extensions, so this involved a lot of reverse-engineering.  This may result in an unorthodox solution, although I did look at other examples.
* The subloop was tricky to implement, but works great and has the desired effect.  For me, the most important thing is that when this extension is disabled, it doesn't have any harmful effects.  I think I took the right steps to ensure there is no problem when it is disabled.
* I do not wish to combine the extensions for usability reasons.  Although they work similarly, the configuration of these are very different.  I think I would consider only releasing "Scramble Prompts" if having both extensions is problematic, rather than combine them.  Maybe this is a better solution.
* I can't use the callback for the info text file because the file is per variation, not per image.

I found a bug when a 0 is entered for certain configurations.  I'll fix the bug and review your feedback in more detail.  Again, thanks for taking the time to review my extensions.

MarcusNyne:
> MarcusNyne maybe

@light-and-ray , are you available on discord?  I'm working in the extensions update code which doesn't seem to work for branches.  I can see you have worked in this area in March.  Can we have a conversation on discord about it?  My username is marcusnyne.

I see dry_run=True on the fetch call .. not sure why this is supposed to work since FETCH_HEAD is only updated on installation.

Searched for you in the webui-dev channel but couldn't find you.

w-e-w:
> > MarcusNyne maybe
> 
> @light-and-ray , are you available on discord? I'm working in the extensions update code which doesn't seem to work for branches. I can see you have worked in this area in March. Can we have a conversation on discord about it? My username is marcusnyne.
> 
> I see dry_run=True on the fetch call .. not sure why this is supposed to work since FETCH_HEAD is only updated on installation.
> 
> Searched for you in the webui-dev channel but couldn't find you.

- https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/15797

MarcusNyne:
@w-e-w, I completed the changes to the extensions.  Here is a comprehensive response to your feedback.  Thank you for the detailed feedback.

**Changes**
- I changed the code structure per your suggestion.  This had other benefits as I was able to keep the prompt file/class name instead of using prompt2.  I like the new organization much better.
- I changed the variation folder to use the directory filename pattern setting.  This was a more elegant solution and respects this setting.  The variation folder is put in a subfolder under this setting.  So, if the setting is likely to change between variations, there will be multiple folders containing variation folders  (I tested with [datetime]).  I don't think this is a common scenario, but the code works predictably.  I believe the default value is [none].  If this is cleared, then the prompt is used, which would change between variation, but again it acts predictably, so I think this is fine.
- I wasn't able to use your suggestion for saving the text file, since there is one file per variation, not per image.  However, I did clean this up by using the image safe path from the processed object instead of trying to calculate the folder path.  This helps with compatibility with the above directory pattern.
- I used your suggestion for 'directories_filename_pattern'.  Although it is interesting that I implemented it this way before I read your comment, so thats why I am accessing it via the data property.
- I changed the enabled indicator to use the InputAccordion.  Thank you for the suggestion.  This works better.  I performed the necessary cleanup of the javascript and made sure it was working properly on the img2img tab.
- I tested img2img thoroughly.  Previously, my focus was on txt2img.

**Other Feedback**
- I decided not to merge the extensions.  They are very different in their use, and I felt that combining them would make them harder to use because the configuration is so different.  Scramble prompts makes a larger change variations, Tweak weights is for fine-tuning.
- I appreciate the feedback on the subloop, and that you took the time to understand it. This is the core of how these extensions work, it required a lot of work and testing to make sure this works flawlessly.  It should not cause any problems when the extension is not enabled, so if someone has a problem, there is an easy remedy.

w-e-w:
I still think the Stable Diffusion: Tweak Weights is working very weirdly as mentioned above but I don't want to care anymore so I'll just let that is an index and let you deal with it

MarcusNyne:
@w-e-w , I wasn't able to reproduce your problem with Tweak Weights.  Can you show me an example of a prompt that will reproduce the problem?

I appreciate your help and would like to fix it if I can reproduce it.

Thanks

PR #320, opened 2024-05-26
create webui-fooocus-prompt-expansion.json

power88:
## Info 
https://github.com/power88/webui-fooocus-prompt-expansion
A simple wrapper for the prompt expansion engine in fooocus. Algorithm mainly written by lllyasviel. I and GPT-4o do some modification to support sd-webui with a simple python file. 

Works well as an extension on stable-diffusion-webui.

## Checklist:

### DIFF
--- a/extensions/webui-fooocus-prompt-expansion.json
+++ b/extensions/webui-fooocus-prompt-expansion.json
@@ -3,8 +3,6 @@
     "url": "https://github.com/power88/webui-fooocus-prompt-expansion.git",
     "description": "a simple wrapper of fooocus prompt expansion engine in stable-diffusion-webui",
     "tags": [
-        "script",
-        "dropdown",
         "prompting"
     ]
 }
--- a/extensions/webui-fooocus-prompt-expansion.json
+++ b/extensions/webui-fooocus-prompt-expansion.json
@@ -1,0 +1,10 @@
+{
+    "name": "webui-foooxus-prompt-expansion",
+    "url": "https://github.com/power88/webui-fooocus-prompt-expansion.git",
+    "description": "a simple wrapper of fooocus prompt expansion engine in stable-diffusion-webui",
+    "tags": [
+        "script",
+        "dropdown",
+        "prompting"
+    ]
+}


w-e-w:
sorry for late response
I look at the code and it works but
but I found that you implement some stuff very weirdly / potentially buggy ways 
when I have time I will try to fix those and send over a fixed PR when I have time

power88:
Ok, thank you for your help! As an artist who is not good at running code, I am indeed not very good at writing aesthetically pleasing code.

w-e-w:
it's not about aesthetics it's about functionality
for instance the way you downloaded model is already extremely weird
where you download model is first off strange and I still at this point cannot figure out why you don't just download the model into your model's directory and instead, download into the shared model directory and then copy files into it
more importantly unless I'm mistaken the directory can potentially be wrong

there's also some issues like you're operating on all prompts which means multiple prompts that's fine
but you're so loaded unload model every single time
for doing most likely the same operation a simple thing to do if you wish to operate it on every individual subprompts is to cash the results and if the inputs are the same then  no new calculation needs to be to performed

other things include
wehy seed limits to 63
code that does nothing

these are only the ones I can remember on hand

by the way do you have a link to the original code from lllyasviel

---

I'm not sure if I'm using the extension wrong but for some reason every image I generated using extension seems to be from Cthulhu

power88:
Okay, I understand the issues you're talking about:

Here is the source file related to the functionality: [https://github.com/lllyasviel/Fooocus/blob/main/extras/expansion.py](https://github.com/lllyasviel/Fooocus/blob/main/extras/expansion.py)

First, regarding the method of downloading the model. At that time, I didn't quite understand and thought that the downloaded models would be stored in a shared directory for unified management, rather than being stored separately in the plugin's folder (for example, sd-webui-controlnet was initially stored in the root directory and later moved to the shared directory).

Secondly, about the frequent unloading and loading of models. I indeed did not consider the need to implement caching to save performance and avoid the models being repeatedly loaded during processing. This issue will also be resolved as soon as possible.

The reason for locking the seed to 63 is that I saw lllyasviel restrict the seed to 63 when calling the function, and I thought exceeding this threshold might cause unexpected issues. (https://github.com/lllyasviel/Fooocus/blob/main/experiments_expansion.py)

Thank you for your guidance. I will resolve these issues as soon as possible.

w-e-w:
if you don't mind waiting a bit 
I should have time to write the fixes over the weekend
> unless you wish to try to fix them yourself of course 

~~actually I already wrote most of the fixes a couple of days ago but I accidentally hard rest the repo and lost the changes 
~~

power88:
Of course! Thank you so much for offering to help optimize the plugin! I really appreciate it since I'm not very good at Python. I'm happy to wait until the weekend for you to write the fixes.

w-e-w:
here you go I think these are the things that I cared to fix
- https://github.com/power88/webui-fooocus-prompt-expansion/pull/1

power88:
Got! Thanks for your pull request!

w-e-w:
I changed the tags a bit
removed `script` `prompting` tag leaveing only `prompting`

PR #323, opened 2024-06-03
Create sd-webui-bilingual-localization-bluelovers.json

bluelovers:
## Info 

Refactored from journey-ad Bilingual translation, rewritten in TypeScript for improved bug fixes and performance enhancements.

## Checklist:

### DIFF
--- a/extensions/sd-webui-bilingual-localization-bluelovers.json
+++ b/extensions/sd-webui-bilingual-localization-bluelovers.json
@@ -1,0 +1,9 @@
+{
+    "name": "Bilingual Localization Refactored Fork",
+    "url": "https://github.com/bluelovers/sd-webui-bilingual-localization.git",
+    "description": "Refactored from journey-ad Bilingual translation, rewritten in TypeScript for improved bug fixes and performance enhancements.",
+    "added": "2024-06-04",
+    "tags": [
+        "UI related"
+    ]
+}
--- a/extensions/sd-webui-bilingual-localization-bluelovers.json
+++ b/extensions/sd-webui-bilingual-localization-bluelovers.json
@@ -2,7 +2,6 @@
     "name": "Bilingual Localization Refactored Fork",
     "url": "https://github.com/bluelovers/sd-webui-bilingual-localization.git",
     "description": "Refactored from journey-ad Bilingual translation, rewritten in TypeScript for improved bug fixes and performance enhancements.",
-    "added": "2024-06-04",
     "tags": [
         "UI related"
     ]


w-e-w:
I don't use the extension myself so I don't know the details but apparently this extension fixes some issues of the original extension which is no longer being maintained or at least inactive for 10 months

is there any reason to not use the fork as the default?
in other words is there a reason to not replace the original entry?

@journey-ad
> still sending a ping even though I don't think he will respond

bluelovers:
i don't want replace original entry in extensions index

w-e-w:
> i don't want replace original entry in extensions index

if that is the case then you must not use the same repo name `sd-webui-bilingual-localization`
as this would cause your extension to be installed on the same directory as journey-ad version

I'm not sure I'm doing something wrong or not but it doesn't seem to be working for me

I installed a localization for zh_TW
set User interface` - `Localization` as None
and set bilingual Bilingual Localization file to zh_TW
as you an see it;s not working 
the file path the is `undefined` 404
image

---

- change the repo name / URL to somethin else
you can add a prefix or something as long as it is unique

also
- don't input `"added": "2024-06-04",` it is automatically generated on merge

make sure it works or either that or maybe I do something wrong

bluelovers:
> but if that is the case then you must Not use the same repo name sd-webui-bilingual-localization
as this would cause your extension to be installed on the same directory as journey-ad version

> change the repo name

will not change, because want to make user can't install both

> undefined` 404

that bug is from original

w-e-w:
https://github.com/bluelovers/sd-webui-bilingual-localization/commit/a648dfbbef51ae28321487e011f56bc5315d2dc2#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR12

I see that you pushing update that says to fix this issue but are you sure it's fixed?
image
it seems to be so getting the same issue on [a648dfbb](https://github.com/bluelovers/sd-webui-bilingual-localization/commit/a648dfbbef51ae28321487e011f56bc5315d2dc2)

---

> will not change, because want to make user can't install both

you do realize that works both ways right
this basically makes it so that and existing user cannot (hard) switch from journey-ad's version to your version

bluelovers:
image

w-e-w:
如果你英文不好用中文 別玩看圖猜謎
[29b5bf07](https://github.com/DominikDoom/a1111-sd-webui-tagcomplete/commit/29b5bf07010a966c6de11c4fde1db98d1dd3972a)
image
image
image

bluelovers:
> 如果你英文不好用中文 別玩看圖猜謎

因為我雖然講中文 但我不論是英文還是中文都跟人有溝通困難XD

看起來只能用傻方法 直接刪除 設定檔內的 bilingual_localization_dirs 讓他在重來一次

w-e-w:
> 看起來只能用傻方法 直接刪除 設定檔內的 bilingual_localization_dirs 讓他在重來一次

可以正常了 算了我不管了 當作通過

PR #326, opened 2024-06-07
Added sd-notes

DocWizard:
## Info 
Repo url: https://github.com/DocWizard/sd-notes-tab

This is something that I feel should be a base function. I've whipped up an extension which adds a new tab for some basic note taking. Full disclosure, this was done with he help of an LLM. I write technical documentation, not python scripts!

Now, i realize this functionality can be replaced with having notepad open in a separate window... But I like having everything in one place. I thought some others might, too. :)

## Checklist:

### DIFF
--- a/extensions/sd-notes-tab.json
+++ b/extensions/sd-notes-tab.json
@@ -1,0 +1,10 @@
+{
+    "name": "Notes Tab",
+    "url": "https://github.com/DocWizard/sd-notes-tab.git",
+    "description": "Adds a simple notes tab.",
+    "tags": [
+        "script",
+        "tab",
+        "UI related",
+    ]
+}


w-e-w:
this is a bit awkward

almost a month ago someone submitted essentially the same extension but due to his implementation being basice and with issues I told him I will make a batter version
https://github.com/w-e-w/sd-webui-notepad

in my version I already implemented `history` `multi-file` `multi-user`

my extension isn't completely finished yet because I wish to change the UI a bit
I'm trying to make it a floating window as supposed to a `tab` or a `accordion`

but I got busy and haven't have time to finalize it

there is a non-exactly policy with the index, in general they should only be one extension that served a very similar purpose
> similar extensions are allowed when the use case / design philosophy is different or is a fork that has significant differences the constitute a different entry

in this case both my extensions and yours are basically doing the same exact thing
so I don't want to put your extension on the index
so unless you made an extension better than mine unfinished extention I'm not going to put your extension on the list

Transparency
normally this only applies to extensions that is already on the index so I'm a bit bending the rules here
if all things went to plan I will have this extension release soonish
Sorry

DocWizard:
Lmao, no problem whatsoever. If you're making a better version, then it's all good, I'll just wait for that. Cheers!

PR #348, opened 2024-07-20
add chara-searcher

NON906:
## Info 

https://github.com/NON906/chara-searcher

"editing" is added to automatically remove the background.
Also, "query" is added to automatically tag images.
If they are not appropriate, please be corresponding.

## Checklist:

### DIFF
--- a/extensions/chara-searcher.json
+++ b/extensions/chara-searcher.json
@@ -1,0 +1,10 @@
+{
+    "name": "chara-searcher",
+    "url": "https://github.com/NON906/chara-searcher.git",
+    "description": "This is a extension for 'character images search' from image and tags.\n- Extract characters from multiple images or videos (background is automatically removed).\n- You can search the extracted character images by image or tags.\n- Search results can be exported and used as is for learning Stable Diffusion's Lora, etc.",
+    "tags": [
+        "tab",
+        "editing",
+        "query"
+    ]
+}


w-e-w:
some issues

---

requirements.txt
```
gradio==3.41.2
```
https://github.com/NON906/chara-searcher/blob/1316b61637b353fa701078d49803450aaf90a832/requirements.txt#L3

I understand that since you provided standalone mode you put all the requirements inside one single requirements.txt

but this would highly likely cause issues especially `gradio==3.41.2`

when we eventually upgraded to [Gradio 4](https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/14184) or if a user is on an old version of where you are that is not using `gradio==3.41.2`, your extentions insulation sequence with Force installation of `gradio==3.41.2` which will break things

you would have to provide separate installation sequences / requirements for webui and standalone model

my PR should fix this issue
- https://github.com/NON906/chara-searcher/pull/1

w-e-w:
I've encountered another issue during installation

but it's time I did not happen again so I'm not sure what the conditions are
`launch.run(f'"{launch.python}" -m mim install mmengine "mmcv>=2.0.0" mmdet')`
this install mmcv
for some reason once it installs mmcv 2.2.0 and once 2.1.0
for me 2.2.0 seems to be incompatible with something (I'm not sure what) if this happens it will cause the extension to not work

NON906:
Sorry for the inconvenience.

The pull request has been reflected.

I also added version specifications for mmcv and mmdet in the following commits.
https://github.com/NON906/chara-searcher/commit/10ce90e908f116c597b32dd2ccef9792f33090a1

PR #352, opened 2024-07-23
Add json for extra-network-side-panel-for-a1111 extension

CurtisDS:
## Info 

https://github.com/CurtisDS/extra-network-side-panel-for-a1111

## Checklist:

### DIFF
--- a/extensions/extra-network-side-panel-for-a1111.json
+++ b/extensions/extra-network-side-panel-for-a1111.json
@@ -1,0 +1,8 @@
+{
+    "name": "extra-network-side-panel-for-a1111",
+    "url": "https://github.com/CurtisDS/extra-network-side-panel-for-a1111.git",
+    "description": "Adds a toggle to move the extra network cards to the side of the screen ",
+    "tags": [
+        "UI related"
+    ]
+}
--- a/extra-network-side-panel-for-a1111.json
+++ b/extra-network-side-panel-for-a1111.json
@@ -1,0 +1,8 @@
+{
+    "name": "extra-network-side-panel-for-a1111",
+    "url": "https://github.com/CurtisDS/extra-network-side-panel-for-a1111.git",
+    "description": "Adds a toggle to move the extra network cards to the side of the screen ",
+    "tags": [
+        "UI related"
+    ]
+}
--- a/extra-network-side-panel-for-a1111.json
+++ b/extra-network-side-panel-for-a1111.json
@@ -1,8 +1,0 @@
-{
-    "name": "extra-network-side-panel-for-a1111",
-    "url": "https://github.com/CurtisDS/extra-network-side-panel-for-a1111.git",
-    "description": "Adds a toggle to move the extra network cards to the side of the screen ",
-    "tags": [
-        "UI related"
-    ]
-}


w-e-w:
> test with current dev / 1.10.0-RC

I don't think the gigantic blank space is intentional
image

initially when I installed extension it seems to work fine but after I switch `Extra Networks directory view style` from `Tree` to `Dirs` the blank the blank space appears
initially I thought it was incompatibility with `Dirs` mode but even after I switch back to `Tree` mode the blank space somehow persist
image

I don't think you want to push this extension to the public at the stage because anyone who experience this is going to have a very bad impression of this extension

---

there's also something that I considered an annoyance
every time I open the side-panel, checkpoint is always the tab selected
it doesn't seem to remember the last selected tab and they don't seem to be a way to change the default selected tab
personally I consider this to be not very convenient

CurtisDS:
image
The only time I could get this to happen is after adjusting the width using this slider. I have checked and it happens only with this extension installed and it doesnt matter if you have toggled the side nav, so there is something wrong there. But I just wanted to confirm with you that you adjusted that slider when you were testing just to make sure this is the cause.

w-e-w:
as I mentioned a book this first happened to me when I switched the mode from `Tree` to `Dirs

and I just retested again in Firefox, this is the first time I opened up with webui in Firefox after installing your extension
> so this should be a "fresh start"

after the web page loaded already in the broken state
if I then adjusted the width slider the dimension "jumps" and is possibly more broken

https://github.com/user-attachments/assets/3f99b28b-04c7-4d5f-bade-48f0e2a22edd

note this video was not from the first time I loaded the web page in Firefox, but I don't think the behavior is different

note I did try clearing caches cookie on Chrome, didn't help

CurtisDS:
I normally run in `Dirs` mode and I only get that bugged layout after adjusting the width of the settings column as described above. I also changed to `Tree` and then back to `Dirs` and couldn't replicate the issue without playing with the width.

I've pushed a change that should fix the layout issue that happens when you adjust the width of the settings, maybe that will solve it. The new commit also adds the ability to remember what last network tab you were on, although its only for the session and doesn't share between txt2img and img2img. (AKA if you're on the lora tab for txt2img, close the panel, switch to img2img and then open the panel... it will pick the checkpoint tab because there was no previously visited tab from the img2img side).

By default it will open the Checkpoint tab if you haven't opened a tab yet. If I were to try and make the default tab a configurable setting Id need to know what extra network tabs are available on the python side. I don't know an easy way of getting that list because extensions can add tabs.

w-e-w:
nice after your fix the bugged layout is fixed for me

I sent over a PR implementing the `default tab`
it is possible you can Implement in some way else using JavaScript but here's me using python

I also found another serious bugs concerning localization
when localization as the innerHTML will be modified comparison doesn't work anymore
this causes the panel to not respond

I send a "partial" int the same PR for the localization issue but it's still partially broken 
considering the amount of people that are using it localization I thinks it's best that localization issue is also fixed before it's merged into the index

CurtisDS:
Thanks for the PRs. I pushed a fix that addresses the issues with localization. It should work now.

w-e-w:
ahhh  thanks I see
https://github.com/CurtisDS/extra-network-side-panel-for-a1111/commit/892c6787d100d8ea78c63008ee6e87518b1ba404#diff-6705926c077b48a0f5508de911f2ca3d7a1902205222ae532fad556cdc74b50fL202-R207

nice working with you

PR #362, opened 2024-08-11
[rq]request add sd-webui-decadetw-auto-messaging-realtime

xlinx:
## Info 

please review: 
https://github.com/xlinx/sd-webui-decadetw-auto-messaging-realtime
thx
## Checklist:

### DIFF
--- a/extensions/sd-webui-decadetw-auto-messaging-realtime.json
+++ b/extensions/sd-webui-decadetw-auto-messaging-realtime.json
@@ -1,0 +1,8 @@
+{
+    "name": "sd-webui-decadetw-auto-messaging-realtime",
+    "url": "https://github.com/xlinx/sd-webui-decadetw-auto-messaging-realtime.git",
+    "description": "Auto messaging you or group to mobile phone about sd-web-ui state in realtime.",
+    "tags": [
+        "script"
+    ]
+}


w-e-w:
is there a reason why you wouldn't just use a remote desktop

xlinx:
Here are some ideas for you to consider. 1. You can watch when u went to; lazy mode. 2. The messages you receive during your IM app chat list which can be as a log like records. 3. Give it a try and see if it feels different! 4. by the way i have many public works keep running, use IM to monitor feels really good, therefor I want share to guys loving sd-webui.

w-e-w:
nope
not convinced in fact less convinced

---

- The basic concept is to monitor using this and edit with remote desktop.

if this is the use case then use the remote desktop it's better in every way

- You can watch it later in a "lazy mode."

great I'm very lazy so I'm going to watch screenshots on the instant message apps it's going to be so entertaining
this is sarcasm

- Messages you receive in your IM app's chat list can serve as a record like logs.

just because it can be used as a log doesn't mean that you should use this as log there are better ways of making a log if you really need to

- You can also use it for auto messaging within chat groups for team work.

just tell me what sort of strange use case with group chat teamwork need screenshots and other information you sent on the regular basis

I'm not going to merge it to the index because I don't endorse it

- Give it a try and see if it feels different! By the way, I have many public projects running, and using IM to monitor them whic is feels really good (relaxing). Therefore, I want to share this with people who love use sd-Webui.

the only time that a screenshot is actually useful is when you're sitting in front of a computer and if you're sitting in front of computer you don't need a screenshot if you need to send a then screenshot then take a screenshot manually when you're aware of the contents

I basically see zero case for your extension
and for those niche situations that it might be useful therefore better ways of implementing it then the way you did it

---

if you truly think this is useful go ahead continue development don't let me stop you
but as of now I can't see that this is a good idea so I'm not going to endorse it and I'm not going to put it on the index
if it turns out that I'm wrong and that this extension does indeed have a wide user base then you can resubmitted to index

xlinx:
okay, thank for the review, I provide usage situstion recently, a art university student render the combine of animdiff in sd-webui, she work this project in the school computers room, that need take care the result about every around 45mins. Mostly school has firewall so the remotedesktop are not work well. and also more and more remote desktop are not free now, include teamview anydesk(5min) realvnc(not free now). Anyway thanks for your review, truly.

xlinx:
This ext provide 1.sd-image 2.sd-text 3. PC-state ex: cpu gpu temp. 4 screenshot will be canceled, which is security issue.
 In fact... when i try to find the best way to capture tempture on linux win os, but not found the good method, so use thescreenshot... I will delete it.

PR #363, opened 2024-08-12
y4my4m-pixelizer

y4my4my4m:
## Info 
https://github.com/y4my4my4m/y4my4m-pixelizer

This is a very simple script that simply shrinks the image down using neareset neighbour by the set pixel amount, with the option to scale it back up. As well as the option to remove the background using the first top-left pixel's color.

Before: 
image

After:
image

## Checklist:

### DIFF
--- a/extensions/y4my4m-pixelizer.json
+++ b/extensions/y4my4m-pixelizer.json
@@ -1,0 +1,10 @@
+{
+    "name": "y4my4m-pixelizer",
+    "url": "https://github.com/y4my4my4m/y4my4m-pixelizer",
+    "description": "A pixelization extension for images that are already pixelart, this will make the image cleaner and you can remove the background.",
+    "tags": [
+        "script",
+        "pixelart",
+        "image"
+    ]
+}


w-e-w:
there's an existing old script that is not currently maintained but as far as I can tell it's still works
if possible I would prefer to update the old extension current UI practices as opposed to add a new extension that does basically the same thing
see https://github.com/Leodotpy/sd-pixel

it does everything yours dose and more accept "remove the background using the first top-left pixel's color"
> seriously, why the top left pixel if you really want to do make it configurable

y4my4my4m:
@w-e-w 
sure, that's doable, should i make a fork of sd-pixel and push an updated version instead or..?
A few people have been complaining about not being able to set the palette colors nor the dithering/quantization in sd-pixel, i guess i could add that too.

Also, top-left pixel is a common thing in spritesheets with the pink background.
Making it customizable is not an issue though

w-e-w:
ahhh

w-e-w:
https://github.com/w-e-w/sd-pixel

w-e-w:
any suggestions before I make a PR

w-e-w:
for future reference the pixelart tag is not a valid tag defined in 
https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/blob/extensions/tags.json
https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/363/files#diff-b8ed784df217098e9d0d9600db267045d95000f04c30838a68bb9552027e8ba1R7

y4my4my4m:
- background transparency with tolerance (customize which pixel to use?)
- allow changing the [dithering](https://github.com/w-e-w/sd-pixel/blob/bd3de8c263c8337ac2015752568bd1f969ab093d/scripts/postprocessing_pixel.py#L56)
- custom quantization? (something [like that](https://github.com/devilismyfriend/PXL8/blob/5538d23beb03b0d3191229149a3ae6d3337b8c65/scripts/pixl8.py#L58) i suppose?)
- force a certain palette (i saw another script was made for that but cant find it right now)

one pixelscript to rule them all

Also, my bad about the tag, should have caught that.

w-e-w:
> Also, top-left pixel is a common thing in spritesheets with the pink background.
Making it customizable is not an issue though

spritesheets  as in basically different animation steps of a pixel art object?

I suppose that's a game industry thing?

y4my4my4m:
@w-e-w yes exactly, these:
image

w-e-w:
> force a certain palette (i saw another script was made for that but cant find it right now)

this?
https://github.com/mrreplicart/sd-webui-pixelart

y4my4my4m:
Might have been this one
https://github.com/C10udburst/stable-diffusion-webui-scripts/tree/master/pixel_art

y4my4my4m:
@w-e-w yeah exactly, as far as I understand it it's more of a legacy standard so that you do not to have to deal with the alpha channel

in the case of AI, a few models are trained on these kind of pixelart and produces a non-transparent image anyway, so reusing the "pink background" method makes sense

w-e-w:
> in the case of AI, a few models are trained on these kind of pixelart and produces a non-transparent image anyway, so reusing the "pink background" method makes sense

---

I feel like transparency could be a standalone extension and

like there are lots of ways you might want to set transparency for example
if you use auto background remover extensions like `rembg`
like what you did to your top left corner or specific color with tolerance
also depends on what you need the output format can be different
obviously the modern stuff we likely would just use Alpha Channel
but maybe in some cases you would like to instead embed transparency color palette metadata (not sure what it's called) in the specific format if supported

y4my4my4m:
@w-e-w it's not a bad idea
the reason why i made this was simply for my particular pixel art making workflow and im doing everything using API calls so a simple call simplified everything

w-e-w:
>  im doing everything using API calls so a simple call simplified everything
wait ... if I my code reading is correct we seem to have neglected support different extra modules in API
if we support this then this shouldn't be an issue in the first place....

---
 
I was testing with different modes of dithering
I'm not sure if I screwed up somewhere because it doesn't seem to make a lot of difference

PR #375, opened 2024-10-12
Pull request for sd-webui-aspect_ratios-dd

zentrocdot:
## Checklist:


w-e-w:
generally we don't add extension if there is something existing that is already too similar as this clutter up the index
and as they already exist similar extensions for setting the aspect ratio I'm hesitant to add this extension but considering that you seems trying to achieve something different I think it's acceptable to be added to index

but then you start adding multiple variations of of the same type
I assuming that you wish to provide different options for different types of workflows for different users and allows the user to choice which type they like, giveing thes no problem user choice but it will clutter the index and sp is a no go

so if you wish to have your extension listed please combine them into one extension

I've written a simple demo extension for a method that can be used by you to create an extension that user decide whitch variations they wish to use
https://github.com/w-e-w/sd_webui_multi_version_extension_demo

w-e-w:
oh so what's up with your tags

as far as I can see your extension is basically a "UI related" extension
you should remove all other tags

PR #379, opened 2024-10-22
sd-webui-quickrecents

MINENEMA:
## Info 

Small UI addition to view recent txt2img/img2img generations and copy their generation parameters
https://github.com/MINENEMA/sd-webui-quickrecents

## Checklist:

### DIFF
--- a/extensions/sd-webui-quickrecents.json
+++ b/extensions/sd-webui-quickrecents.json
@@ -1,0 +1,8 @@
+{
+    "name": "sd-webui-quickrecents",
+    "url": "https://github.com/MINENEMA/sd-webui-quickrecents.git",
+    "description": "View recent generations and get their generation parameters straight from the txt2img/img2img tabs",
+    "tags": [
+        "UI related"
+    ]
+}


w-e-w:
I reviewed extension and found it to actually not be working on A1111

so I decided to make a fix PR
- https://github.com/MINENEMA/sd-webui-quickrecents/pull/1

for details read the PR

you can provide your own implementation of the fixes or use my PR after that this can be merge to the index

MINENEMA:
Oh wow thanks for the detailed breakdown, 

initially made it only looking at how the PNG info tab sending image params to other tabs worked and yeah i was stumped on the gradio cache issue so i just used the first working solution i came across

i assumed it worked for automatic1111 cause of its similar implementation of the `parameter_copypaste.register_paste_params_button` it was using

Ill merge your fix and add some minor changes
Thanks!

w-e-w:
Ability to only paste certain parameters (ex: prompt only)

I forgot to mention
Settings > Infotext > Disregard fields from pasted infotext

basically web UI has a built-in function to discard specific variables if you wish

the only issue that it's "global" at least if you're using the built-in function for parsing and passing the arguments

if you don't mind that it's Global then you can just use that but if you want specific control for your
but if you want to only discard different keys when you're pasting through your extension then you would have to be creative or maybe Implement your own functions
it might be possible to patch it so that you can detect that you are extension is activating the pasting
and when doing so apply a different set of discards

w-e-w:
the reason why it didn't work was because they `gr.SelectData` that you were using to get the path of the image does not 
in the the version fo gradio of a1111

PR #390, opened 2024-11-28
Added stable-diffusion-webui-chat-gpt-prompts.json

ilian6806:
## Info 

https://github.com/ilian6806/stable-diffusion-webui-chat-gpt-prompts

* Generate SD prompts variations with ChatGPT
* Optionally translate the prompt from another language with DeepL
* Optionally use a custom ChatGPT model
* Optionally use custom ChatGPT instructions
* UI state is preserved across page reloads

## Checklist:

### DIFF
--- a/extensions/stable-diffusion-webui-chat-gpt-prompts.json
+++ b/extensions/stable-diffusion-webui-chat-gpt-prompts.json
@@ -1,7 +1,7 @@
 {
     "name": "stable-diffusion-webui-chat-gpt-prompts",
     "url": "https://github.com/ilian6806/stable-diffusion-webui-chat-gpt-prompts",
-    "description": "Generates SD prompt variations with ChatGPT and DeepL",
+    "description": "Generates SD prompt variations with ChatGPT and DeepL. Requires an OpenAI API key (paid subscription) and optionally a DeepL API key.",
     "tags": [
         "prompting",
         "online"
--- a/extensions/stable-diffusion-webui-chat-gpt-prompts.json
+++ b/extensions/stable-diffusion-webui-chat-gpt-prompts.json
@@ -1,0 +1,9 @@
+{
+    "name": "stable-diffusion-webui-chat-gpt-prompts",
+    "url": "https://github.com/ilian6806/stable-diffusion-webui-chat-gpt-prompts",
+    "description": "Generates SD prompt variations with ChatGPT and DeepL",
+    "tags": [
+        "prompting",
+        "online"
+    ]
+}


w-e-w:
sorry for the late review being rather busy recently

---

issues with current extension

there's no have no documentation / warning about that the user needs to have a API key to open AI to use ChatGPT
there's also no instructions on how to get the API key

since you have no instructions about API keys if the user install your extension and press your `✨` button
**Nothing will happen the pop-up won't appear** as if your extension is broken

this is because you have a line that would throw an error if the user has not assigned an API key yet
https://github.com/ilian6806/stable-diffusion-webui-chat-gpt-prompts/blob/bccb030ab640aa562b57ae2eee6f229e112a0d49/javascript/gptp.core.js#L227

this converter will happen on the first install if the user hasn't pressed `apply setting` whitch
the config file doesn't have the entry the value you're trying to reference is undefined

and there is no need of adding an additional API endpoint `/gptp/config.json` for getting settings
since you endpoint is also not authenticated, and your put the API keys in settings, your literary exposing your API key to anyone
please remove the custom api because it is unnecessary and for security reasons

you can simply access any setting key using JavaScript by using the global variable `opts`
for example `opts.gptp_deepl_api_key`

addition advantage of using `opts`
currently if you update your API key you need to reload your web page in order for it to work
because I think you only fetch the settings once during `init()`
if you use `opts` the value will be automatically updated without needing a reloading

---

summary
1. at sufficient documentation and warning about OpenAI subscription needed and about API key
2. remove the unsecure API end point and use opts
3. make sure it works on a new install

---

since I do not have a OpenAI subscription, I cannot truly test this extension myself
so I would like you to provide a short video of it woring (preferably from a clean install)

ilian6806:
Hi @w-e-w! 

Thanks for your feedback, really appreciated! I didn't know we have a global opts variable.

I made the following changes:
- Removed config API route. Replaced with ops global variable.
- Created wiki pages with docs for getting OpenAI and DeepL API keys
- Added paid subscription info in the extension description
- Added a popup when we don't have an OpenAI API key in settings with a link to the wiki pages 
- Created video with extension usage
- Added YouTube link to the video in the README

Demo video -> [https://www.youtube.com/watch?v=zz_pok7smjY](https://www.youtube.com/watch?v=zz_pok7smjY)

w-e-w:
didn't test myself but based on your videos looks like it works fine

just one suggestion about the video
disable grammarly while making the video
someone that doesn't know what grammarly is may think it's a part of your extension

PR #391, opened 2024-12-13
Create Gelbooru-Prompt-Randomizer.json

RED1cat:
## Info 

https://github.com/RED1cat/Gelbooru-Prompt-Randomizer
## Checklist:

### DIFF
--- a/extensions/Gelbooru-Prompt-Randomizer.json
+++ b/extensions/Gelbooru-Prompt-Randomizer.json
@@ -4,6 +4,6 @@
     "description": "Finds a random post from Gelbooru and takes tags from it according to filters.",
     "tags": [
         "prompting",
-        "online",
+        "online"
     ]
 }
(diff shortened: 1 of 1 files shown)


w-e-w:
didn't found any particular issues with the extensions itself

but you entry has an extra comma at the end of the tags which makes the json invalid 
> I fixed it

---

this is just a pure suggestion, decide yourself if you think these are good ideas

currently your extension workflow is, click `Randomize` button and the current prompt will be replaced
while this works if you just want random and nothing else it doesn't play well with other aspects of the us such as `lora`
as every time you randomize the current prompt is replaced
not to mention I think it's quite possible that users have some extra prompts themselves that they may wish to add in
and after every  `Randomize` the user has to scroll up and down the page in order to see the new random prompt, this gets tedious quite quickly

I would suggest that you add a custom text box output for the random generated prompts
and let the user decide if they actually wish to use the prompt in their current prompts

and adding couple of method of applying to current prompts as opposed to "just replace"
such as
- append
- prepend
- and possibly even advanced ones such as keeping extra networks the same or something not too sure how useful this will be

at the end of the day if the random prompts are inside a different text box they can always copy and modify them themselves to suit their needs as opposed to having all everything done automatically

it is totally possible that there are people that find automatically applying and replacing current prompt is useful
in this case you can always keep it as an option

RED1cat:
This is my first experience writing an extension, so it turned out to be a bit primitive design. Thanks for the ideas for improving the extension, I'll try to use them in the next update.

PR #393, opened 2024-12-14
Add Perchance Extension

ouoertheo:
## Info 
https://github.com/ouoertheo/sd-webui-perchance.git

## Checklist:

### DIFF
--- a/extensions/perchance.json
+++ b/extensions/perchance.json
@@ -1,0 +1,9 @@
+{
+    "name": "Perchance",
+    "url": "https://github.com/ouoertheo/sd-webui-perchance.git",
+    "description": "Insert text from Perchance.org generators. Requires NodeJS (sorry, no way around that).",
+    "tags": [
+        "script",
+        "prompting"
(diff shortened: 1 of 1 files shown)


w-e-w:
I think I understand what's your extension is trying to achieve
if I understand it correctly it's basically downloading the text generators html from perchanc and running them in nodejs to get the outputs and having a nodejs server that connects it with web UI
and the reason you want this is so that people can access generators that hav already been created by the perchance community

---

unfortunately I have to bring you some bad news
we think that running nodejs it's not a good idea
especially when the input is unknown and can be anything
the obfuscated javascript code from perchanc also makes people uncomfortable 

due to these concerns we cannot list this extension on the index

---

well the use cases a bit different users can achieve similar functionality by using extensions such as `dynamic prompts` or `wildcard`

ouoertheo:
Hey, no sweat, thanks for considering it.

w-e-w:
thanks for understanding

